PR #1, opened 2016-02-01
Update apt cache before attempting to install keepalived package

Logan2211:


evrardjp:
Hello, thanks for the pull request.

I don't see a reason to accept this, as this is not following ansible best practices, and will lengthen every keepalived role run.

Instead, I think it's maybe better to add something to your playbook like:
pre_tasks:
  apt: update_cache=yes

Thank you for the PR anyway!

PR #13, opened 2016-05-09
fix example usage in README.md

mancdaz:
Both keepalived_servers and keepalived_hosts was being used. This
commit fixes both examples to use keepalived_hosts to aid clarity.


evrardjp:
FYI, this is just an example, and doesn't reflex what's being done on specific consuming projects.
But thanks for the fixing the consistency issue!

PR #14, opened 2016-07-27
Support instance notification scripts.

timl:
Allows instance notification scripts to be defined.  Uses the same dict format as the sync group notification scripts.


evrardjp:
Sorry for the delay.

Looks good to me.

PR #17, opened 2016-09-08
create redhat.yml

mgariepy:
in order to be able to include the role even when the conditional is false, the Gather variables needs to pass.


evrardjp:
Thanks for your contribution!

PR #21, opened 2016-10-20
added optional track_interface option

jardleex:
Hello @evrardjp,

here comes a small PR which adds keepalived's **track_interface** option to the role.

Best

Jard


evrardjp:
Looks great!

evrardjp:
Included in version 2.2.0 of the role.

PR #24, opened 2016-11-02
Fix linting issues for ansible-lint 3.4.1

Logan2211:
Update role to pass latest ansible-lint tests.

Tying the UCA apt key status to the keepalived package state is a fair change because we do not need to enforce the UCA key state if we are not forcing updates on keepalived.


evrardjp:
Rechecking for the idempotency issues.

PR #34, opened 2017-01-30
Add SELinux module for keepalived's ping cmd

major:
The `ping` command used by keepalived causes SELinux AVC denials on
CentOS 7. This causes keepalived to think that all of the systems
are unstable and it can't find a suitable host to run services.

This patch adds a policy that allows keepalived to function without
causing AVCs.

For more details:
  https://bugs.launchpad.net/openstack-ansible/+bug/1660426

### DIFF
--- a/tasks/keepalived_selinux.yml
+++ b/tasks/keepalived_selinux.yml
@@ -1,0 +1,66 @@
+---
+# Copyright 2017, Major Hayden <major@mhtx.net>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Get list of SELinux modules loaded
+  command: semodule -l
+  changed_when: False
+  register: selinux_modules
+
+- name: Ensure SELinux packages are installed
+  yum:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - libselinux
+    - libselinux-devel
+  when:
+    - '"keepalived_ping" not in selinux_modules.stdout'
+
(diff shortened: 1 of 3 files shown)


tasks/keepalived_selinux.yml
```diff
@@ -0,0 +1,51 @@
...
+    src: keepalived_ping.te
+    dest: /tmp/ansible-keepalived-selinux/keepalived_ping.te
+    owner: root
+    group: root
+    mode: "0755"
+
+- name: Compile and load SELinux module
+  command: "{{ item }}"
+  args:
+    creates: /etc/selinux/targeted/active/modules/400/keepalived_osa/cil
```
evrardjp: I'd prefer removing the _osa if possible.
I'd be enclined to use _ping instead.

tasks/keepalived_selinux.yml
```diff
@@ -0,0 +1,51 @@
...
+    chdir: /tmp/ansible-keepalived-selinux
+  with_items:
```
evrardjp: If one item succeeds and the second fail, another run will not re-run the full list of commands, correct?
Shouldn't there be a {{ item }} in the created file?

tasks/keepalived_selinux.yml
```diff
@@ -0,0 +1,51 @@
...
+    - checkmodule -M -m -o keepalived_ping.mod keepalived_ping.te
+    - semodule_package -o keepalived_ping.pp -m keepalived_ping.mod
+    - semodule -i keepalived_ping.pp
+
+- name: Remove temporary directory
```
evrardjp: I'm glad SELinux is not in this role gate, it would never have been able to merge: This breaks idempotency.
Isn't there a way to check and then only run the folder creation/destruction when necessary?
Or should we write a module instead?

tasks/keepalived_selinux.yml
major: Darn -- OSA shouldn't be in there anyway. Typo.

tasks/keepalived_selinux.yml
major: These commands are okay to re-run. The last one (`semodule -i keepalived_ping.pp`) is the only one which takes more than a second to finish.

tasks/keepalived_selinux.yml
major: Good point. I can put in a check up front.

evrardjp:
Thanks major!

PR #35, opened 2017-02-01
Allow for parameter templating in keepalived

jwitko:
This PR allows for the templating of /etc/sysconfig/keepalived parameters.  The template will create /etc/sysconfig/keepalived with the specified parameters.  By default the role will create the file exactly as it exists from the generic package installation.


jwitko:
@evrardjp All checks have passed so please merge at your convenience.

evrardjp:
It seems generally ok.

I'll have to add a release note and/or docs, and I will try manually for 16.04, because I don't have that gated (yet).

It may take a while, but I'll try to find some time this week-end.

evrardjp:
I didn't find some time this week-end, I will try this evening. Sorry for the inconvenience.

evrardjp:
Still in process.

defaults/main.yml
```diff
@@ -37,4 +37,5 @@ uca_apt_repo_url: "http://ubuntu-cloud.archive.canonical.com/ubuntu"
 
 keepalived_instances: []
 keepalived_sync_groups: {}
-keepalived_bind_on_non_local: False
\ No newline at end of file
+keepalived_bind_on_non_local: False
+keepalived_options: "-D"
```
evrardjp: I'd rather have this defaulted to nothing. This way we don't change standard behavior.

tasks/main.yml
```diff
@@ -58,6 +58,16 @@
   notify:
     - restart keepalived
 
+- name: Configure /etc/sysconfig/keepalived parameters
+  template:
+    src: sysconfig_keepalived.j2
+    dest: "{{ keepalived_sysconfig_file_path }}"
```
evrardjp: This is likely not enough:
1) The ubuntu init script won't load the generated file in defaults under ubuntu 14.04
2) The ubuntu init script crashes when giving this under ubuntu 16.04
3) The centos works.

Please adapt to restrict to the appropriate OS, or fix it under the other OSes, please

templates/sysconfig_keepalived.j2
```diff
@@ -0,0 +1,16 @@
...
+# --vrrp               -P    Only run with VRRP subsystem.
+# --check              -C    Only run with Health-checker subsystem.
+# --dont-release-vrrp  -V    Dont remove VRRP VIPs & VROUTEs on daemon stop.
+# --dont-release-ipvs  -I    Dont remove IPVS topology on daemon stop.
+# --dump-conf          -d    Dump the configuration data.
+# --log-detail         -D    Detailed log messages.
+# --log-facility       -S    0-7 Set local syslog facility (default=LOG_DAEMON)
+#
+
+KEEPALIVED_OPTIONS="{{ keepalived_options }}"
```
evrardjp: This breaks ubuntu 16.04.

evrardjp:
Hello @jwitko , I've updated this role, and tested under the classical 2 ubuntu + centos.
I have a full functional test (you can try at home with a simple ``vagrant up``) that works. I don't think this PR is needed anymore. You could re-open if it's needed.

PR #38, opened 2017-03-14
Adding timeout attribute for check script

BjoernT:
This fix adds the timeout attribute for check scripts to allow execution beyond 0seconds, which is particularly important for ping checks with used more than one ICMP packet to verify connectivity


BjoernT:
Apparently setting rise, timeout, fall as string will lead to a bug in keepalived (for trusty at least).

evrardjp:
Fair point for both.

Sadly, I can't accept these as is. Could you have 2 PR instead?
- one for the bug fixing with rise, timeout, and fall (this will be marked as minor).
- the other one, will be marked as major, should implement the new feature to add timeout attribute for check script. By default this should be set to 2. It should be optional like it is in keepalived docs.

Do you go ahead or I do these two?

evrardjp:
You can probably change many things in the meantime with openstack-ansible.

BjoernT:
@evrardjp OK I reduced this PR to just adding the attribute

evrardjp:
This is now included in the 2.6.0 release of the keepalived role, based on your feedback!
I'm bumping this in osa master and ocata release (with a release note).

Thanks a lot!

PR #40, opened 2017-03-20
Fix parsing issues for the rise, fall and interval configuration

BjoernT:
Passing the check script configuration as strings will
cause keepalived, at least version 1.2.19, to parse them
as boolean causing race conditions like check scripts
always failing when the response take longer than one second
and faulting the keepalived states.


evrardjp:
Thanks!

PR #42, opened 2017-03-28
Support for virtual routes for VRRP instances

Ray76:
Extended template to support virtual routes for VRRP, adding routes if becoming master, deleting them for backup.


evrardjp:
Looks good. Use at your own risk!

PR #43, opened 2017-04-11
set missing value after warmup key

jardleex:
Hello @evrardjp,

here comes a small fix on the warmup setting on the misc_check's.
According to the official [documentation](https://github.com/acassen/keepalived/blob/master/doc/keepalived.conf.SYNOPSIS#L688) a integer value needs to be provided after the keyword.

Best

Jard


evrardjp:
Good catch!

evrardjp:
Do you want me to tag this, or are you using master?

jardleex:
A tag would be nice as we use themto use a fixed version in our environment.
Thank you!

evrardjp:
Ok doing it.

PR #47, opened 2017-07-03
SELinux: allow keepalived to cat haproxy pid

major:
On CentOS 7, an SELinux denial occurs when keepalived tries to read
haproxy's PID file. This patch adds SELinux policy to allow this
activity and breaks the SELinux compilation part of the role into
a reusable set of tasks.

### DIFF
--- a/tasks/keepalived_selinux.yml
+++ b/tasks/keepalived_selinux.yml
@@ -31,8 +31,6 @@
     - '"keepalived_ping" not in selinux_modules.stdout'
 
 - include: keepalived_selinux_compile.yml
-  with_items:
-    - "keepalived_haproxy_pid_file"
-    - "keepalived_ping"
+  with_items: "{{ keepalived_selinux_compile_rules }}"
   loop_control:
     loop_var: selinux_policy_name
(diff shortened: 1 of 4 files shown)


evrardjp:
I will spin up my centos tests today.

evrardjp:
My tests don't do integration tests of keepalived + haproxy, merely keepalived.
This PR passed my upgrade tests. Doing the greenfield ones right now.

tasks/keepalived_selinux.yml
```diff
@@ -30,39 +30,9 @@
...
-  when:
-    - '"keepalived_ping" not in selinux_modules.stdout'
-
-- name: Remove temporary directory
-  file:
-    path: /tmp/ansible-keepalived-selinux/
-    state: absent
-  when:
-    - '"keepalived_ping" not in selinux_modules.stdout'
+    - "keepalived_haproxy_pid_file"
```
evrardjp: Therefore, I think these items should be coming from a variable, named keepalived_selinux_policy_files, with a default value of either [] or [ 'keepalived_ping'].

I can't adapt your PR, so could you please update it?

This variable could then be updated in OSA (with a group var for example), to compile the files.
OSA can provide additional files, like the haproxy pid one.

evrardjp:
The code passed my tests, but I'd rather see some of these features be optional, by either using a role as dependency, or making these SELinux bits running optionally depending on a var (see review above)

PR #49, opened 2017-08-02
remove the use of iteritems

cloudnull:
iteritems is a py2 callable and will break when using py3.

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>


evrardjp:
It looks good, and passes the automatic tests I have here. I need to do a few tests manually before merging and tagging.

evrardjp:
Passed my vagrant tests for interop. If there is a bug, then it's the fault of my tests!

PR #52, opened 2017-12-15
Ensure keepalived starts after networking

major:
This patch adds a systemd override that makes keepalived start once
all networking interfaces are fully online.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -178,3 +178,22 @@
   service:
     name: "{{ keepalived_service_name }}"
     enabled: "yes"
+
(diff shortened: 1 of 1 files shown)


evrardjp:
This doesn't seem to pass ansible 2.2 checks under 14.04. I think this should be done conditionally on ansible_sys_mgr being systemd?

major:
@evrardjp This updated patch should be better.

evrardjp:
Thanks I will have a look asap.

evrardjp:
Works!

PR #53, opened 2017-12-21
Mask systemd service for Ubuntu 16.04 during install

### LINKED ISSUE #18: virtual IP not coming up
Hi, 

i have been testing stable/newton and one thing I find is that the vip does not come up (I test on 3 controllers). This makes the repo build fail during setup-infrastructure , and the solution is to manually login to all the 3 controllers and do a keepalive restart. This makes one of them working and can continue with the rest of the process.

james-portman:
Previous method of disabling keepalived from starting during install did
not work
Fix #18 in Virtualbox/Vagrant

### DIFF
--- a/tasks/keepalived_install.yml
+++ b/tasks/keepalived_install.yml
@@ -59,6 +59,19 @@
   when:
     - ansible_os_family | lower == 'debian'
     - check_if_present | changed
+    - prevent_start_mask_systemd is not defined
+  tags:
+    - keepalived-prevent-start
+
+- name: Prevent keepalived from starting on install - systemd mask
+  file:
+    state: link
+    src: "/dev/null"
+    dest: "{{ prevent_start_file }}"
+  when:
+    - ansible_os_family | lower == 'debian'
+    - check_if_present | changed
+    - prevent_start_mask_systemd is defined
   tags:
     - keepalived-prevent-start
 
@@ -81,3 +94,9 @@
   tags:
     - keepalived-config
     - keepalived-prevent-start
+
+- name: "Reload systemd if we deleted service mask"
+  systemd:
+    daemon_reload: yes
+  when:
+    - prevent_start_mask_systemd is defined
--- a/tasks/keepalived_install.yml
+++ b/tasks/keepalived_install.yml
@@ -59,7 +59,8 @@
   when:
     - ansible_os_family | lower == 'debian'
     - check_if_present | changed
-    - prevent_start_mask_systemd is not defined
+    - ansible_service_mgr is defined
+    - ansible_service_mgr != 'systemd'
   tags:
     - keepalived-prevent-start
 
@@ -71,7 +72,8 @@
   when:
     - ansible_os_family | lower == 'debian'
     - check_if_present | changed
-    - prevent_start_mask_systemd is defined
+    - ansible_service_mgr is defined
+    - ansible_service_mgr == 'systemd'
   tags:
     - keepalived-prevent-start
 
@@ -83,6 +85,7 @@
     cache_valid_time: "{{ (ansible_pkg_mgr == 'apt') | ternary(cache_timeout, omit) }}"
   tags:
     - keepalived-apt-packages
+  notify: "systemctl-daemon-reload"
 
 - name: Revert keepalived start prevention
   file:
@@ -94,9 +97,3 @@
   tags:
     - keepalived-config
     - keepalived-prevent-start
-
-- name: "Reload systemd if we deleted service mask"
-  systemd:
-    daemon_reload: yes
-  when:
-    - prevent_start_mask_systemd is defined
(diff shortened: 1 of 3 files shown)


evrardjp:
I am currently on holiday and will test that as soon as I am back to work. Seems ok at first sight, but I want to do a full series of manual tests. Sorry for the inconvenience.

james-portman:
OK thank you, and no problem!

evrardjp:
I am back this week and will investigate further. Sorry for the delay!

tasks/keepalived_install.yml
```diff
@@ -59,6 +59,19 @@
   when:
     - ansible_os_family | lower == 'debian'
     - check_if_present | changed
+    - prevent_start_mask_systemd is not defined
```
evrardjp: I'd rather not define another variable, and use ansible_service_mgr instead.
so I'd adapt the condition here by:
```
when:
  - ansible_os_family | lower == 'debian'
  - check_if_present | changed
  - ansible_service_mgr is defined
  - ansible_service_mgr != 'systemd'
```

tasks/keepalived_install.yml
```diff
@@ -59,6 +59,19 @@
...
+
+- name: Prevent keepalived from starting on install - systemd mask
+  file:
+    state: link
+    src: "/dev/null"
+    dest: "{{ prevent_start_file }}"
+  when:
+    - ansible_os_family | lower == 'debian'
+    - check_if_present | changed
+    - prevent_start_mask_systemd is defined
```
evrardjp: I'd change this last condition to:
```
- ansible_service_mgr is defined
- ansible_service_mgr == 'systemd'
```

tasks/keepalived_install.yml
```diff
@@ -81,3 +94,9 @@
   tags:
     - keepalived-config
     - keepalived-prevent-start
+
+- name: "Reload systemd if we deleted service mask"
```
evrardjp: This should probably be a handler. When the package is changed, we should do a systemd daemon_reload.

On top of that it won't work with ansible < 2.4. It only passed my automatic testing because I test with ansible 2.3 and ubuntu 14.

vars/ubuntu-16.04.yml
```diff
@@ -29,5 +29,5 @@ keepalived_uca_repo_dist: "{{ ansible_lsb.codename }}-updates/{{ keepalived_uca_
 keepalived_uca_openstack_release: ocata
 
 ## Prevent start on install
-prevent_start_file: "/usr/sbin/policy-rc.d"
-prevent_start_file_content: "exit 101"
+prevent_start_file: "/etc/systemd/system/keepalived.service"
```
evrardjp: We can keep the prevent_start_file as /etc/systemd/system/keepalived.service, but we should probably remove the variable prevent_start_mask_systemd.

We can keep the variable prevent_start_file_content, and replace its value to false, or omit, which would trigger an error if we are entering the loop incorrectly.

james-portman:
I can have a go at the changed at some point if I get time,
if you have any spare time or want to - then definitely have a go yourself, it's up to you

james-portman:
Ok I’ll try and have a go tomorrow morning if I can!
And thanks!

tasks/keepalived_install.yml
james-portman: Done

tasks/keepalived_install.yml
james-portman: Done

james-portman:
OK just updated, let me know if you want me to squash the commits or any other changes!

evrardjp:
Thanks for the update!

evrardjp:
My first series of tests work. Will ensure the second batch is ok, and I will issue a new tag when the other pending PR is fully tested and working.

tasks/keepalived_install.yml
```diff
@@ -70,6 +85,7 @@
     cache_valid_time: "{{ (ansible_pkg_mgr == 'apt') | ternary(cache_timeout, omit) }}"
   tags:
     - keepalived-apt-packages
+  notify: "systemctl-daemon-reload"
 
 - name: Revert keepalived start prevention
```
evrardjp: We should probably change this task:

For non systemd, the prevent start file should be absent, but for systemd the link should be restored to the service.

evrardjp: My bad, it's something else, some sources ship with wrong files, and load sample files. I will submit a combined patch.

evrardjp:
Your code is now merged! I've created a PR with your commit, added a few extra changes needed to test it properly, and fixed the issue that appeared with the extra test.

Thanks for your contribution!

james-portman:
Thanks! I have another potential one next, will make a PR and see what you think...

evrardjp:
Thank you in advance then! I look forward to the PR.

PR #54, opened 2018-01-04
Improve robustness of Vagrant tests

evrardjp:
The ansible_play_host is not ordered the same way at every run.
If the first node happen to be keepalived2 instead of 1 or 3, the
VRRP convergence test would fail, as the test was looking into
the presence of the VRRP master ip on the eth1 NIC, which is not
the NIC name for ubuntu 16.04.

This moves to use {{ vrrp_nic }} instead of hardcoding eth1, and
ensures as much as possible a deterministic order for ansible 2.4
runs.

On top of that, a vagrant provision done twice would report
the provisioning as "changed" instead of "ok", only because we
are using a raw task for installing python2 on the ubuntu 16.04
node, which is considered as the minimum to run ansible.
This task to install python2 shouldn't be
considered as a change of ansible-keepalived itself, and shouldn't
impact idempotency tests.

### DIFF
--- a/tests/deploy.yml
+++ b/tests/deploy.yml
@@ -2,9 +2,13 @@
 - hosts: keepalived2
   gather_facts: no
   tasks:
-    - raw: sudo apt-get install -y python
+    - name: Install python2
+      raw: sudo apt-get update && sudo apt install -y python
+      changed_when: false
 
 - hosts: all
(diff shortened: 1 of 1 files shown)


evrardjp:
This works for ansible 2.2, 2.3 and 2.4 (keeping in mind the comment), so I merge it.

PR #65, opened 2018-09-28
Added ability to specify daemon launch arguments.

noonedeadpunk:
It may be used for providing extra args for keepalived daemon.
With these args it's possible to change logging options or enabling SNMP checks.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -55,3 +55,15 @@ keepalived_bind_on_non_local: False
 # Example:
 #keepalived_global_defs:
 #  - enable_script_security
+
+# Set location of keepalived daemon options file path
+# For Debian based systems it's usually /etc/default/keepalived
+# For RedHat based systems it's usually /etc/sysconfig/keepalived
+keepalived_daemon_options_file_path: ""
+
+# Overriding keepalived daemon extra arguments, which will be applied inside of
+# the keepalived_daemon_options_file_path variable. 
+# Example:
+#keepalived_daemon_default_options_overrides: 
+#  - "DAEMON_ARGS='--snmp'"
+keepalived_daemon_default_options_overrides: []
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -59,7 +59,7 @@ keepalived_bind_on_non_local: False
 # Set location of keepalived daemon options file path
 # For Debian based systems it's usually /etc/default/keepalived
 # For RedHat based systems it's usually /etc/sysconfig/keepalived
-keepalived_daemon_options_file_path: ""
+keepalived_daemon_options_file_path: "{{ _keepalived_daemon_options_file_path }}"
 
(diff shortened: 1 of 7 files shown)


noonedeadpunk:
It's a strange thing on semaphore, as Ubuntu 14.04 definitely has this file by default:
```
root@test-ubuntu-keepalived:~# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.5 LTS
Release:        14.04
Codename:       trusty
root@test-ubuntu-keepalived:~# cat /etc/default/keepalived
# Options to pass to keepalived

# DAEMON_ARGS are appended to the keepalived command-line
DAEMON_ARGS=""
root@test-ubuntu-keepalived:~#
```

evrardjp:
Of course, I am okay with an alternative implementation that passes testing.

Keep in mind you could also edit this AFTER the fact in your own play, but that kinda defeats the point.

tasks/main.yml
```diff
@@ -54,6 +54,21 @@
...
     - restart keepalived
 
+- name: Configure keepalived extra params
+  lineinfile:
+    line: "{{ item }}"
+    regexp: "^{{ item.split('=')[0] }}"
+    dest: "{{ keepalived_daemon_options_file_path }}"
+    state: present
+  with_items: "{{ keepalived_daemon_default_options_overrides }}" 
+  when:
```
evrardjp: as said above, the when is always true, due to the definition of both variables in defaults.

defaults/main.yml
```diff
@@ -55,3 +55,15 @@ keepalived_bind_on_non_local: False
 # Example:
 #keepalived_global_defs:
 #  - enable_script_security
+
+# Set location of keepalived daemon options file path
+# For Debian based systems it's usually /etc/default/keepalived
+# For RedHat based systems it's usually /etc/sysconfig/keepalived
+keepalived_daemon_options_file_path: ""
```
evrardjp: if keepalived_daemon_options_file_path is defined here, its condition in tasks is always true.

evrardjp: keepalived_daemon_options_file_path seems distro specific, and should be moved to vars, providing a good "default"

tasks/main.yml
```diff
@@ -54,6 +54,21 @@
...
+    - keepalived_daemon_default_options_overrides is defined
+    - keepalived_daemon_options_file_path is defined
```
evrardjp: if you move keepalived_daemon_options_file_path to vars/, as said above, both will always be true.

Because the list "keepalived_daemon_default_options_overrides" is empty by default [], this task would do nothing, which is fine, so you could remove the conditionals.

evrardjp:
Doing manual testing right now.

evrardjp:
Passed manual testing.

PR #68, opened 2018-11-06
Extend realservers check configuration

touchardv:
Make it possible to use `TCP_CHECK`s in the keepalived configuration.


evrardjp:
@touchardv Hello. I am not using this feature myself, and therefore it won't be properly tested neither automatically or manually. I am fine with merging this, but it's up to you, you're on your own there ;)

PR #69, opened 2018-11-28
add many VIP s and real servers

sunnoy:


evrardjp:
Still a very big PR, and not much info provided in there. Could we make multiple PRs, each targetting a part of the code?

Thank you.

PR #87, opened 2019-03-21
Add ssl_get check

dwerder:
Add the SSL_GET section for real_servers, as described in
https://manpages.debian.org/stretch/keepalived/keepalived.conf.5.en.html
Search for "SSL_GET"

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -167,6 +167,19 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
     }
     {% endfor %}
     {% endif %}
+    {% if rserver.ssl_get is defined %}
+    {% for sslcheck in rserver.ssl_get %}
+    SSL_GET {
+      url {
+        path {{ sslcheck.url_path }}
+        digest {{ sslcheck.url_digest }} 
+      }
+      connect_timeout {{ sslcheck.connect_timeout | default('3') }}
+      nb_get_retry {{ sslcheck.nb_get_retry | default('3') }}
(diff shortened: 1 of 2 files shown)


templates/keepalived.conf.j2
```diff
@@ -167,6 +167,19 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
...
     {% endif %}
+    {% if rserver.ssl_get is defined %}
+    {% for sslcheck in rserver.ssl_get %}
+    SSL_GET {
+      url {
+        path {{ sslcheck.url_path }}
+        digest {{ sslcheck.url_digest }} 
+      }
+      connect_timeout {{ sslcheck.connect_timeout | default('3') }}
+      nb_get_retry {{ sslcheck.nb_get_retry | default('3') }}
```
evrardjp: Is that not simply retry? Should we add retry? I haven't found this in documentation.

dwerder: Thats how its cocumented in the official man page:

https://manpages.debian.org/stretch/keepalived/keepalived.conf.5.en.html

And also the exemple files, of the keepalived RPM show that options. So I copied them and the config works.

dwerder: You can also watch here, search for SSL_GET example sections:
http://www.keepalived.org/LVS-NAT-Keepalived-HOWTO.html

evrardjp: I checked another doc, keepalived.org too. I guess we can go ahead, it's enough proof for me :)

evrardjp:
Do you want me to retag, or are you consuming from master?

PR #88, opened 2019-04-09
Added weight to the template to make adjustment of priority possible

styleart:
Just found out that there was no `weight` option for the keepalived configuration file and added it. Cheers.

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -50,7 +50,9 @@ vrrp_script {{ name }} {
   interval {{ details.interval | default(5) }}   # checking every {{ details.interval | default(5) }} seconds (default: 5 seconds)
   fall {{ details.fall | default(3) }}           # require {{ details.fall | default(3) }} failures for KO (default: 3)
   rise {{ details.rise | default(6) }}           # require {{ details.rise | default(6) }} successes for OK (default: 6)
-  weight {{ details.weight | default(0) }}       # adjust priority by this weight
+  {% if details.weight is defined %}
+  weight {{ details.weight | default(0) }}       # adjust priority by this weight, (default: 0)
+  {% endif %}
   {% if details.timeout is defined %}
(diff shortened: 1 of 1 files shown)


templates/keepalived.conf.j2
```diff
@@ -50,6 +50,7 @@ vrrp_script {{ name }} {
   interval {{ details.interval | default(5) }}   # checking every {{ details.interval | default(5) }} seconds (default: 5 seconds)
   fall {{ details.fall | default(3) }}           # require {{ details.fall | default(3) }} failures for KO (default: 3)
   rise {{ details.rise | default(6) }}           # require {{ details.rise | default(6) }} successes for OK (default: 6)
+  weight {{ details.weight | default(0) }}       # adjust priority by this weight
```
evrardjp: Good catch, we never defined it!

The code is good: the default matches the standard defaults.

However, I would prefer if this was wrapped with `{% if details.weight is defined %}`.

Without this conditional, running this role in an existing production environment would result in the addition of "weight 0" in the template, and restart keepalived. This could cause downtime if the role is not run in a serial way (yes, I know, I am VERY careful...).

Could you add conditionals and test if the template doesn't change if weight is not provided please?

Thank you in advance.

styleart: Good point! I wasn't thinking about the restart of keepalived and possible outcomes.

Added if statement and tested it - no restart if no `weight` was defined 👍

evrardjp:
Very busy day today.

I will try to have time for this today, if not don't hesitate to ping me again tomorrow :)

This looks good enough to warrant a new Y bump (x.y.z) after merging/tagging.

evrardjp:
I will be testing this next week.

styleart:
No problem, it's not urgent for us :)

evrardjp:
Testing today.

evrardjp:
Tagged, release produced, and pushed to galaxy. Thanks for your patch!

PR #102, opened 2019-07-13
Pre-commit fixups

retr0h:
I use gilt[1] to pull in my Ansible roles, and a series of
pre-commit hooks which correct trailing whitespace and
newlines.  These are triggered when I vendor your project
into my repo.  Since this doesn't look too controversial,
figured I'd submit a PR.

[1]: https://gilt.readthedocs.io/en/latest/

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,4 +199,3 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-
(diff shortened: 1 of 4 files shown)


evrardjp:
I will have a more detailed look on monday, but I am not against the idea :)

retr0h:
> It won't impact any production deploy using openstack-ansible too, so it's even possible to mark the next release as a bugfix.

Sounds good... Shout out to openstack-ansible, @major and @cloudnull too :P

evrardjp:
Haha, even if those ppl are "has-been" for OSA (like me!), I will convey your regards to those ppl :)

Have a good day!

PR #111, opened 2019-09-02
Add virtual_rules to template

miloaec:
Add the possibility to add virtual_rules in keepalived.conf


evrardjp:
We have to check if this doesn't mean an unnecessary restart of the service for ppl which don't ask for virtual_rules. (due to whitespaces)

PR #119, opened 2019-10-23
add ipv6.ip_nonlocal_bind, group track_scripts

MartinVerges:
add ipv6.ip_nonlocal_bind
add per group track_scripts
make script user configurable


evrardjp:
Thanks for the patch!

It might be impactful for some people, and therefore I made this a new minor version bump (3.6.0).

You should be able to use it from galaxy ansible today.

PR #120, opened 2019-12-20
SELinux fixes (check mode)

jacekkow:
Currently in check mode (--check) role fails with error:

> The conditional check 'selinux_policy_name not in selinux_modules.stdout' failed. The error was: error while evaluating conditional (selinux_policy_name not in selinux_modules.stdout): Unable to look up a name or access an attribute in template string ({% if selinux_policy_name not in selinux_modules.stdout %} True {% else %} False {% endif %}).
> Make sure your variable name does not contain invalid characters like '-': argument of type 'StrictUndefined' is not iterable

These commits prevent such error and installing SELinux-related packages when list `keepalived_selinux_compile_rules` is empty.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -30,6 +30,7 @@
 
 - include: keepalived_selinux.yml
   when:
+    - keepalived_selinux_compile_rules | length > 0
     - ansible_selinux.status is defined
     - ansible_selinux.status == "enabled"
   tags:
--- a/tasks/keepalived_selinux.yml
+++ b/tasks/keepalived_selinux.yml
@@ -17,6 +17,7 @@
   command: semodule -l
   changed_when: False
   register: selinux_modules
+  check_mode: no
 
 - name: Ensure SELinux packages are installed
   yum:
--- a/tasks/keepalived_selinux.yml
+++ b/tasks/keepalived_selinux.yml
@@ -31,6 +31,8 @@
     - '"keepalived_ping" not in selinux_modules.stdout'
 
 - include: keepalived_selinux_compile.yml
+  when:
+    - selinux_policy_name not in selinux_modules.stdout
   with_items: "{{ keepalived_selinux_compile_rules }}"
   loop_control:
     loop_var: selinux_policy_name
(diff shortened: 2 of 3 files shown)


tasks/main.yml
```diff
@@ -30,6 +30,7 @@
 
 - include: keepalived_selinux.yml
   when:
+    - keepalived_selinux_compile_rules
```
evrardjp: that is always not falsy by default. Could you explain why do we need this? Maybe we could move this to ``| length > 0 `` for those making the list an empty list, if that's what you want to achieve? (I think it has a clearer meaning when reading the code)

jacekkow: I know it is not empty by default. Nevertheless I do not want to have any additional SELinux packages (i.e. libselinux-devel) installed when this list is empty (or possibly False/None/...).

I am OK with any of:
- `keepalived_selinux_compile_rules`
- `not not keepalived_selinux_compile_rules`
- `keepalived_selinux_compile_rules | bool`
- `keepalived_selinux_compile_rules | length > 0`

Choose the one you prefer and I'll update this PR.

jacekkow:
There are two commits for two separate issues:
- in check mode selinux_modules is not registered causing the role to fail (logic),
- SELinux packages are installed even when there are no modules to install (bloat).

The first issue (check mode) is not eliminated completely as task "Compile and load SELinux module" still fails with default settings due to nonexistent `/tmp/ansible-keepalived-selinux-{{ selinux_policy_name }}`, but was enough for me at the time (I set `keepalived_selinux_compile_rules: []`). Later I have encountered the second issue and added another commit with `when:` clause.

By the way, I see that all tasks in `keepalived_selinux_compile.yml` have the same `when:` clause that can be factored out to `keepalived_selinux.yml`.

tasks/main.yml
evrardjp: ``keepalived_selinux_compile_rules | length > 0`` would be my preference.

If the list is non-empty, it would be running, and you can still override with an empty list (with []) to skip in the when condition.

evrardjp:
> By the way, I see that all tasks in `keepalived_selinux_compile.yml` have the same `when:` clause that can be factored out to `keepalived_selinux.yml`.

Totally correct :)
I used to do this differently for bug reasons in very old ansible. Thanks for the reminder on the need to cleanup the cruft there! :)

evrardjp:
Awesome!

PR #121, opened 2019-12-20
Do not pass the copyright notice to the generated file

jacekkow:
While the template itself is copyright-protected, I would not consider the generated config file a Derivative Work, therefore (as far as I know) it should not contain the Apache license template.

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -1,4 +1,5 @@
 #jinja2: lstrip_blocks: True
+{#
 # Copyright 2015, Jean-Philippe Evrard <jean-philippe@evrard.me>
 #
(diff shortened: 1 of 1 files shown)


evrardjp:
Are you in a hurry to merge this? Because I will need to sync with people, and it's holiday period right now.

jacekkow:
Take your time. I also see a possibility of adding an option whether to add this template or not (default set to not impact existing users).

evrardjp:
I don't want to bloat the role of variables :)

At first sight, people are okay with it. Let's merge this after the holiday period... I will tag the other two patches in a release before that if you're in hurry.

My responses are relatively slow nowadays, because I am on vacation myself, sorry for the delays.

PR #122, opened 2019-12-20
Allow not to create overrides for keepalived.service (systemd)

jacekkow:
Some operating systems may have proper settings already - i.e. CentOS 7+

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -206,7 +206,9 @@
   file:
     path: /etc/systemd/system/keepalived.service.d/
     state: directory
-  when: ansible_service_mgr == 'systemd'
+  when:
+    - ansible_service_mgr == 'systemd'
+    - keepalived_systemd_overrides | bool
 
 - name: Apply keepalived override to start after network is up
   ini_file:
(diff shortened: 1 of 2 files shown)


tasks/main.yml
```diff
@@ -206,7 +206,9 @@
   file:
     path: /etc/systemd/system/keepalived.service.d/
     state: directory
-  when: ansible_service_mgr == 'systemd'
+  when:
+    - ansible_service_mgr == 'systemd'
+    - keepalived_systemd_overrides
```
evrardjp: could you add "| bool" , so that keepalived_systemd_overrides gets validated, please?

tasks/main.yml
```diff
@@ -219,4 +221,14 @@
   with_items:
     - 'Wants'
     - 'After'
-  when: ansible_service_mgr == 'systemd'
+  when:
+    - ansible_service_mgr == 'systemd'
+    - keepalived_systemd_overrides
```
evrardjp: ditto

tasks/main.yml
```diff
@@ -219,4 +221,14 @@
...
+
+- name: Remove keepalived overrides
+  file:
+    path: /etc/systemd/system/keepalived.service.d/override.conf
+    state: absent
+  when:
+    - ansible_service_mgr == 'systemd'
+    - not keepalived_systemd_overrides
```
evrardjp: not (keepalived_systemd_overrides | bool)

evrardjp:
I will tag the role with a new version when your series of patches have merged, if you don't mind.

PR #125, opened 2020-01-09
Extend and add Healthchecks

thehawkes:
Aloha,

added more variables for:
TCP_CHECK
SSL_GET
MISC_CHECK

added additional checks
HTTP_GET
DNS_CHECK

https://manpages.debian.org/stretch/keepalived/keepalived.conf.5.en.html

Regards.

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -170,15 +170,17 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
     {% if rserver.tcp_checks is defined %}
     {% for tcp_check in rserver.tcp_checks %}
     TCP_CHECK {
+      connect_port {{ tcp_check.connect_port }}
+      connect_timeout {{ tcp_check.connect_timeout | default('5') }}
       {% if tcp_check.connect_ip is defined and tcp_check.connect_ip %}
       connect_ip {{ tcp_check.connect_ip }}
       {% endif %}
-      {% if tcp_check.connect_port is defined and tcp_check.connect_port %}
-      connect_port {{ tcp_check.connect_port }}
+      {% if tcp_check.retry is defined and tcp_check.retry | int %}
+      retry {{ tcp_check.retry }}
       {% endif %}
-      connect_timeout {{ tcp_check.connect_timeout | default('5') }}
-      retry {{ tcp_check.retry | default('1') }}
-      delay_before_retry {{ tcp_check.delay_before_retry | default('1') }}
+      {% if tcp_check.delay_before_retry is defined and tcp_check.delay_before_retry | int %}
+      delay_before_retry {{ tcp_check.delay_before_retry }}
+      {% endif %}
       {% if tcp_check.bindto is defined and tcp_check.bindto %}
       bindto {{ tcp_check.bindto }}
       {% endif %}
@@ -222,12 +224,12 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
         path {{ http_check.url_path }}
         digest {{ http_check.url_digest }}
       } 
-      {% if http_check.connect_port is defined and http_check.connect_port %}
-      connect_port {{ http_check.connect_port }}
-      {% endif %}
       connect_timeout {{ http_check.connect_timeout | default('3') }}
       nb_get_retry {{ http_check.nb_get_retry | default('3') }}
       delay_before_retry {{ http_check.delay_before_retry | default('2') }}
+      {% if http_check.connect_port is defined and http_check.connect_port %}
+      connect_port {{ http_check.connect_port }}
+      {% endif %}
       {% if http_check.connect_ip is defined and http_check.connect_ip %}
       connect_ip {{ http_check.connect_ip }}
       {% endif %}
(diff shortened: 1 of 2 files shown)


templates/keepalived.conf.j2
```diff
@@ -168,39 +168,133 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
   {% for rserver in vserver.real_servers %}
   real_server {{ rserver.ip }} {{ rserver.port }} {
     {% if rserver.tcp_checks is defined %}
-    {% for tcp_check in rserver.tcp_checks %}
-    TCP_CHECK {
-      connect_port {{ tcp_check.connect_port }}
-      connect_timeout {{ tcp_check.connect_timeout | default('5') }}
-    }
-    {% endfor %}
+      {% for tcp_check in rserver.tcp_checks %}
```
evrardjp: I am fine with this, however, there seems to be an indentation difference. People using tcp_checks will see a diff, which will basically restart their keepalived, even if there is no functional change for them...

While the template looks cleaner this way, I think the file will contain extra spaces. Don't you think we should keep the same indentation as right now in the generated file?

tests/keepalived_haproxy_master_example.yml
```diff
@@ -128,10 +128,74 @@ keepalived_instances:
...
+#              user: 'root'
+#              group: 'root'
 #        ssl_get:
 #          - url_path: '/'
 #            url_digest: 'a797b47875d8fd5c949066182902099d'
 #            # Optional
 #            connect_timeout: 3
 #            nb_get_retry: 3
 #            delay_before_retry: 2
+#            # Ip to connect to, if not set real_server ip will be used
```
evrardjp: Could you add `(Optional)` or `Optional,` at the beggining of the comment, please? That would be awesome.

templates/keepalived.conf.j2
thehawkes: yes there will be an extra space in the config file. i've just added the extra spaces during debugging of a jinja template error. i'll change that back to the default spacing.

tests/keepalived_haproxy_master_example.yml
thehawkes: sure. i'll check my comments again and update it accordingly

templates/keepalived.conf.j2
evrardjp: Awesome, thanks!

tests/keepalived_haproxy_master_example.yml
evrardjp: Perfect!

templates/keepalived.conf.j2
```diff
@@ -170,8 +170,27 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
...
     {% for tcp_check in rserver.tcp_checks %}
     TCP_CHECK {
+      {% if tcp_check.connect_ip is defined and tcp_check.connect_ip %}
+      connect_ip {{ tcp_check.connect_ip }}
+      {% endif %}
+      {% if tcp_check.connect_port is defined and tcp_check.connect_port %}
       connect_port {{ tcp_check.connect_port }}
+      {% endif %}
       connect_timeout {{ tcp_check.connect_timeout | default('5') }}
+      retry {{ tcp_check.retry | default('1') }}
```
evrardjp: Is retry necessary nowadays? I thought it wasn't. In that case I would prefer if we wrap it with a condition. It allows us to not vendor-in a default which could effectively change upstream.

evrardjp:
Thanks for your contribution, it's good quality!

templates/keepalived.conf.j2
thehawkes: at least I've used it in the past :relaxed:
but I think it's a good idea to wrap these optional settings.
I will do that for all my newly added settings, but leave out the current available ones like
https://github.com/evrardjp/ansible-keepalived/blob/9e7a079abca9b0e2381f10e0dab4a0d66f5e8b35/templates/keepalived.conf.j2#L173-L174

https://github.com/evrardjp/ansible-keepalived/blob/9e7a079abca9b0e2381f10e0dab4a0d66f5e8b35/templates/keepalived.conf.j2#L199-L201

because of https://github.com/evrardjp/ansible-keepalived/pull/125#discussion_r365107536

evrardjp: Thanks! That's indeed what I would hope.
Sorry for being a little bit picky, but I do what I think is for everyone's best :)

thehawkes: NP. your code, your rules ;)

templates/keepalived.conf.j2
```diff
@@ -271,12 +273,10 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
     }
     {% endfor %}
     {% endif %}
-    {% if rserver.dns_check is defined %}
+    {% if rserver.dns_checks is defined %}
     {% for dns_check in rserver.dns_checks %}
     DNS_CHECK {
-      {% if dns_check.connect_ip is defined and dns_check.connect_ip %}
       connect_ip {{ dns_check.connect_ip }}
```
thehawkes: i've made the connect_ip option mandatory even if it is optional for keepalived.
just to make sure that there is a value in that dict to prevent ansible dict errors.

evrardjp:
This will be released today, along with another patch.

PR #127, opened 2020-01-15
add advert_int, weight and thresholds

thehawkes:
Hello again,

some smaller additions. 
- vrrp advert_int 
- real_server weight
- real_server uthreshold
- real_server lthreshold

another pull request will follow, because I would like to add `virtual_server_group` 
> This feature offers a way to simplify your configuration by factorizing virtual server
> definitions. If you need to define a bunch of virtual servers with exactly the same
> real server topology then this feature will make your configuration much more readable
> and will optimize healthchecking task by only spawning one healthchecking where
> multiple virtual server declaration will spawn a dedicated healthchecker for every
> real server which will waste system resources.

that would include a split up of the current `keepalived.conf.j2` template to reduce duplicate code without breaking the backward compatibility

Regards

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -84,7 +84,7 @@ vrrp_instance {{ name }} {
   {% if instance.preempt_delay is defined %}
   preempt_delay {{ instance.preempt_delay }}       # Seconds after startup until preemption. 0 (default) to 1,000.
   {% endif %}
-  {% if instance.advert_int is defined and instance.advert_int | int %}
+  {% if instance.advert_int is defined %}
   advert_int {{ instance.advert_int }}
   {% endif %}
   {% if instance.authentication_password is defined %}
@@ -170,14 +170,14 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
 
   {% for rserver in vserver.real_servers %}
   real_server {{ rserver.ip }} {{ rserver.port }} {
-    {% if rserver.weight is defined and rserver.weight | int %}
-    weight {{ rserver.weight }}
+    {% if rserver.weight is defined %}
+    weight {{ rserver.weight | int }}
     {% endif %}
-    {% if rserver.uthreshold is defined and rserver.uthreshold | int %}
-    uthreshold {{ rserver.uthreshold }}
+    {% if rserver.uthreshold is defined %}
+    uthreshold {{ rserver.uthreshold | int }}
     {% endif %}
-    {% if rserver.lthreshold is defined and rserver.lthreshold | int %}
-    lthreshold {{ rserver.lthreshold }}
+    {% if rserver.lthreshold is defined %}
+    lthreshold {{ rserver.lthreshold | int }}
     {% endif %}
     {% if rserver.tcp_checks is defined %}
     {% for tcp_check in rserver.tcp_checks %}
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -84,6 +84,9 @@ vrrp_instance {{ name }} {
   {% if instance.preempt_delay is defined %}
   preempt_delay {{ instance.preempt_delay }}       # Seconds after startup until preemption. 0 (default) to 1,000.
   {% endif %}
+  {% if instance.advert_int is defined and instance.advert_int | int %}
+  advert_int {{ instance.advert_int }}
+  {% endif %}
   {% if instance.authentication_password is defined %}
   authentication {
     auth_type PASS
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -167,6 +167,15 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
 
   {% for rserver in vserver.real_servers %}
   real_server {{ rserver.ip }} {{ rserver.port }} {
+    {% if rserver.weight is defined and rserver.weight | int %}
+    weight {{ rserver.weight }}
+    {% endif %}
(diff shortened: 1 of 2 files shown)


templates/keepalived.conf.j2
```diff
@@ -84,6 +84,9 @@ vrrp_instance {{ name }} {
   {% if instance.preempt_delay is defined %}
   preempt_delay {{ instance.preempt_delay }}       # Seconds after startup until preemption. 0 (default) to 1,000.
   {% endif %}
+  {% if instance.advert_int is defined and instance.advert_int | int %}
```
evrardjp: anything real number between 0 and 1 will be ignored, as it will be floored to 0. I am not sure if that's you want, as the value will basically not be set.

I would rather have an explicit failure by removing ``and instance.advert_int | int`` in the condition.

templates/keepalived.conf.j2
```diff
@@ -167,6 +170,15 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
 
   {% for rserver in vserver.real_servers %}
   real_server {{ rserver.ip }} {{ rserver.port }} {
+    {% if rserver.weight is defined and rserver.weight | int %}
+    weight {{ rserver.weight }}
```
evrardjp: weight being an integer, you can force it here in L174 by adding | int.

templates/keepalived.conf.j2
```diff
@@ -167,6 +170,15 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
...
+    {% endif %}
+    {% if rserver.uthreshold is defined and rserver.uthreshold | int %}
+    uthreshold {{ rserver.uthreshold }}
```
evrardjp: same as weight.

templates/keepalived.conf.j2
```diff
@@ -167,6 +170,15 @@ virtual_server {{ vserver.ip }} {{ vserver.port }} {
...
+    {% endif %}
+    {% if rserver.lthreshold is defined and rserver.lthreshold | int %}
+    lthreshold {{ rserver.lthreshold }}
```
evrardjp: same as weight.

templates/keepalived.conf.j2
thehawkes: That option will simply be not set if it's not an INT and keepalived will use the default `advert_int 1`. 
But I can change that. no problem.

templates/keepalived.conf.j2
thehawkes: But it's already "forced" to be an INT in L173
Do you want these options to fail if it's not INT like the `advert_int` ?

templates/keepalived.conf.j2
thehawkes: alright it's just an converter with default behavior and it will not fail in that stage

https://jinja.palletsprojects.com/en/2.10.x/templates/#int
>Convert the value into an integer. If the conversion doesn’t work it will return 0. 

for example:
with the following template
```    {% if rserver.weight is defined %}
    weight {{ rserver.weight | int }}
    {% endif %}
    {% if rserver.uthreshold is defined %}
    uthreshold {{ rserver.uthreshold | int }}
    {% endif %}
    {% if rserver.lthreshold is defined %}
    lthreshold {{ rserver.lthreshold | int }}
    {% endif %}```

```     - ip: 10.0.0.1
        port: 5445
        weight: 0.2
        uthreshold: blubber
        lthreshold: 12
        tcp_checks:
          - connect_port: 5445
            connect_timeout: 5```

will result in 
```  real_server 10.0.0.1 5445 {
    weight 0
    uthreshold 0
    lthreshold 0
    TCP_CHECK {
      connect_port 5445
      connect_timeout 5
    }```

not really what we want.
the same results with the `advert_int`
so I would suggest to keep the templates as is, with `and instance.advert_int | int` and simply not set it if it's not a correct INT and use the default upstream settings instead.
Same for `weight`, `uthreshold` and `lthreshold`
We could set it to other default values for example `| int(1)` but that stuff needs to be updated if upstream is changing it.

If you would like to fail if it's not an INT, it should be done in `main.yml` ```- name: Configure keepalived
  template:
    src: keepalived.conf.j2
    dest: "{{ keepalived_config_file_path }}"
  tags:
    - keepalived-config
  notify:
    - restart keepalived```
either as an extra task or via `when`
Opinions?

evrardjp: I thought the int in advert_int stands for interval, which didn't need to be an integer. In this case, anything works, including for example 0.82.

So therefore the condition on line 87 should not test or cast into an int (and therefore should be ``{% if instance.advert_int is defined %}``)

The advert_int line 88 is right the way it is right now (``advert_int {{ instance.advert_int }}``).

templates/keepalived.conf.j2
evrardjp: It's not "forced" to be an int on L173.

In fact, we should probably remove the ``and rserver.weight | int`` part of the condition, as it is allowed to have 0 as weight.

However, I think we should have `weight {{ rserver.weight | int }}`  at L174, to force the templating of an integer.

evrardjp:
Thanks for your contribution !

thehawkes:
> Thanks for your contribution !

You are welcome :) 
Thank you for the Role, I just need to add some stuff in order to move our current keepalived setups to ansible.
I've updated the template according to our discussion

evrardjp:
@thehawkes FYI I am not using real_server myself, and it's not tested in CI. 
If we want to improve the testing there, maybe we should discuss it.

Let me propose another version on galaxy.

PR #132, opened 2020-01-30
Add optional virtual_ipaddress_excluded section

tridde:
Keepalived has a feature to include IP addresses that keepalived
should bring up and down, but not include in VRRP packets.


evrardjp:
I am currently in a middle of a move, so it will take a little extra time to review this, sorry for the inconvenience.

evrardjp:
Got the chance to test today. Works!

PR #124, opened 2020-01-09
add support for EL8

pugnacity:
This should add support for EL8 and close #123

PR #135, opened 2020-02-10
Fix for #124

pugnacity:
my changes for #124 where incomplete

### DIFF
--- a/tasks/keepalived_selinux.yml
+++ b/tasks/keepalived_selinux.yml
@@ -25,7 +25,6 @@
     state: present
   when:
     - '"keepalived_ping" not in selinux_modules.stdout'
-    - keepalived_selinux_packages is defined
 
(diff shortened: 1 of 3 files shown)


tasks/keepalived_selinux.yml
```diff
@@ -21,14 +21,11 @@
...
-    name:
-      - libselinux
-      - libselinux-devel
-      - checkpolicy
-      - policycoreutils-python
+    name: "{{ keepalived_selinux_packages }}"
     state: present
   when:
     - '"keepalived_ping" not in selinux_modules.stdout'
+    - keepalived_selinux_packages is defined
```
evrardjp: You can remove that condition. `keepalived_selinux_packages` is defined everywhere.

evrardjp:
Tagged and imported as 3.11.1. Thanks for your contribution @pugnacity !

PR #140, opened 2020-05-06
Add support for `vrrp_track_file` and `track_files`

jjungnickel:


templates/keepalived.conf.j2
```diff
@@ -72,6 +79,18 @@ vrrp_script {{ name }} {
 {% endfor %}
 {% endif %}
 
+{% if keepalived_track_files is defined %}
+{% for name, details in keepalived_track_files.items() %}
+vrrp_track_file {{ name }} {
+  file "{{ details.file }}"
+  weight {{ details.weight | default(1) }} # multiply with weight in file (default: 1, i.e. literal value from file)
+  {% if details.init_file | default(false) %}
+  init_file{% if details.init_file_value is defined %} {{ details.init_file_value }}{% endif %}{% if details.init_file_overwrite | default(false) %} overwrite{% endif %}
```
evrardjp, line 88: I would prefer if you add `| bool` to all those `default(false)` conditionals. If it's defined and truth-y, it won't necessary work. It's a more Ansible-y way to add the ` | bool` to verify the input is as expected. This applies to L87 and L88.

PR #141, opened 2020-06-11
Ubuntu 20.04

jostmart:
Using PPA on Ubuntu 20.04 did fail. Shifted to keepalived_ubuntu_src: "native" since the OS release come with keepalived 2.19 which was OK for my project. Since afaik 2.20 is the newest, I did not spend more time to get near the bleading edge :)


evrardjp:
I am back from holidays!

Could you squash those commits into one? It looks okay to me at first sight.

evrardjp:
Please reopen if necessary :)

PR #145, opened 2020-07-14
Remove install python-apt task: trust target is properly set up for ansible.

fbourqui:
based on ansible_python_interpreter of target

### DIFF
--- a/tasks/keepalived_install.yml
+++ b/tasks/keepalived_install.yml
@@ -45,9 +45,9 @@
   tags:
     - keepalived-repo
 
-- name: Ensure python apt is installed
+- name: Ensure python-apt or python3-apt is installed
   apt:
-    name: python-apt
+    name: "python{{ ansible_python_interpreter | default('') | regex_search('3$') }}-apt"
     state: present
   when:
     - ansible_pkg_mgr == 'apt'


fbourqui:
For me removing the task would make sense, but if you have user that use really old ansible, at least with the fix, we will not end-up with a full python2 stack for nothing on all release with python3.

evrardjp:
Let's remove the task and trust the users to do the right thing.

fbourqui:
Hello,



do you have a use case that is causing any issue when loading the config file once at the beginning  ?

I'm happy to have a look at it.



Cheers,

Fred

 



---- On Fri, 17 Jul 2020 11:17:03 +0200 Jean-Philippe Evrard <notifications@github.com> wrote ----




Let's remove the task and trust the users to do the right thing.

evrardjp:
I have never seen an issue with loading the config file in group_vars or in a variable file. But I am not sure how is that relevant to this PR. Should we discuss this in an issue?

fbourqui:
Sorry,



I did reply to the wrong topic, I had another pull request in another repo.

It's Friday, ..



---- On Fri, 17 Jul 2020 17:42:04 +0200 Jean-Philippe Evrard <notifications@github.com> wrote ----




I have never seen an issue with loading the config file in group_vars or in a variable file. But I am not sure how is that relevant to this PR. Should we discuss this in an issue?

evrardjp:
Sorry for the late answer, while it was Friday, it was certainly friday before my staycation! :)

Do you need further help on getting it done?

tasks/keepalived_install.yml
```diff
@@ -45,9 +45,9 @@
   tags:
     - keepalived-repo
 
-- name: Ensure python apt is installed
+- name: Ensure python-apt or python3-apt is installed
   apt:
```
evrardjp, line 49: maybe, for clarity, we could do a variable, `packagename:` whose content can be
`{{ ansible_python['version']['major'] == 3 | ternary('python3-apt','python-apt') }}`, or something similar?

evrardjp:
My PR, #157 should cover this.

PR #152, opened 2020-08-17
Fix "file was found when using first_found"

in0rdr:
When the role is used alongside some other tasks in a playbook, Ansible tries to include the OS vars but cannot find them.

Let's assume, our playbook is built as follows:
```yaml
- hosts: haproxy
  tags: keepalived
  become: yes
  roles:
    - { role: keepalived }

- hosts: all
  tags: some-other-tasks
  become: yes
  tasks:
    - debug: msg=test
```

The role resides in the directory `roles/haproxy`. However, the following error occurs with the current implementation:
```
No file was found when using first_found. Use errors='ignore' to allow this task to be skipped if no files are found
```
Due to the `always` flag, this is the case even when the role should not be executed in the first place.

As a workaround, one can define the "some-other-tasks" in a separate playbook. This solution works, but I propose to improve the role in a way that you can use it alongside "some-other-tasks" unrelated to keepalived. 

With the fix from this PR, Ansible always finds the correct files on the path `{{ role_path }}/vars"` and continues with an empty file list (skip=true) if no files are found: https://docs.ansible.com/ansible/latest/plugins/lookup/first_found.html

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -16,11 +16,15 @@
 - name: Gather variables for each operating system
   include_vars: "{{ item }}"
   with_first_found:
-    - "{{ ansible_distribution | lower }}-{{ ansible_distribution_version | lower }}.yml"
-    - "{{ ansible_distribution | lower }}-{{ ansible_distribution_major_version | lower }}.yml"
-    - "{{ ansible_os_family | lower }}-{{ ansible_distribution_major_version | lower }}.yml"
-    - "{{ ansible_distribution | lower }}.yml"
-    - "{{ ansible_os_family | lower }}.yml"
+    - files:
+      - "{{ ansible_distribution | lower }}-{{ ansible_distribution_version | lower }}.yml"
+      - "{{ ansible_distribution | lower }}-{{ ansible_distribution_major_version | lower }}.yml"
+      - "{{ ansible_os_family | lower }}-{{ ansible_distribution_major_version | lower }}.yml"
+      - "{{ ansible_distribution | lower }}.yml"
+      - "{{ ansible_os_family | lower }}.yml"
+    - paths:
+      - "{{ role_path }}/vars"
+      skip: true
   tags:
     - always
 


evrardjp:
I am against the idea of skipping if no var is found, because it will silently ignore real failures.
Let me check the rest :)

evrardjp:
So here, for me, you're relying on ansible bad behaviour of loading the improper files.

Can you make sure keepalived is in `roles`, similarly to your `haproxy` role, instead of having keepalived as a subfolder of haproxy?

tasks/main.yml
```diff
@@ -16,11 +16,15 @@
...
-    - "{{ ansible_os_family | lower }}.yml"
+    - files:
+      - "{{ ansible_distribution | lower }}-{{ ansible_distribution_version | lower }}.yml"
+      - "{{ ansible_distribution | lower }}-{{ ansible_distribution_major_version | lower }}.yml"
+      - "{{ ansible_os_family | lower }}-{{ ansible_distribution_major_version | lower }}.yml"
+      - "{{ ansible_distribution | lower }}.yml"
+      - "{{ ansible_os_family | lower }}.yml"
+    - paths:
+      - "{{ role_path }}/vars"
+      skip: true
```
evrardjp, line 27: I am not against everything up to this point.

I would prefer if skip: true is removed.

We should also clarify in meta: the version for which this will run. I am not sure when role_path was introduced.

evrardjp: Seems we are okay there for the role_path.

in0rdr: The `skip: true` seems to be necessary to run the tasks tagged "some-other-tasks" in the same playbook (I guess that the `always` tag is also a bit invasie?).

Alternative 1 (works):

I could split the monolithic playbook in several small playbooks to go around that issue (one playbook for the role, another playbook for the tasks/snippets). The PR is irrelevant in that case.


Alternative 2 (does not work with tasks in playbook?):

I test on centos machines, so the `redhat.yml` is the relevant file. If it were evident, in which directory the `vars/redhat.yml` file should be placed, I could also just add an empty file in that directory as another workaround. The file would exist, even though it will never be used because the playbook should just execute the tasks tagged "some-other-tasks". However, I tried to place a file in almost every place in the file system. Seems like Ansible does not load the files from any `vars` directory for executing the tasks tagged "some-other-tasks"?

```bash
ansible-playbook -i dev/inventory playbook.yml -t some-other-tasks --check
```

evrardjp: what if you do -t always,some-other-tasks ?

evrardjp: it shouldn't be that hard....

evrardjp: > The `skip: true` seems to be necessary to run the tasks tagged "some-other-tasks" in the same playbook (I guess that the `always` tag is also a bit invasive?).

the idea of the `always` tag was that it always runs, regardless of the limitation of tags. So that _should_ be fine in here.
We've done this in the `openstack-ansible` project consuming this role.

> 
> Alternative 1 (works):
> 
> I could split the monolithic playbook in several small playbooks to go around that issue (one playbook for the role, another playbook for the tasks/snippets). The PR is irrelevant in that case.

I would say it's probably the best, but I would like us to find out what's wrong first, because what you seem to want to do _does not seem_ to be terrible, UX wise.

> 
> Alternative 2 (does not work with tasks in playbook?):
> 
> I test on centos machines, so the `redhat.yml` is the relevant file. If it were evident, in which directory the `vars/redhat.yml` file should be placed, I could also just add an empty file in that directory as another workaround. The file would exist, even though it will never be used because the playbook should just execute the tasks tagged "some-other-tasks". However, I tried to place a file in almost every place in the file system. Seems like Ansible does not load the files from any `vars` directory for executing the tasks tagged "some-other-tasks"?
> 
> ```shell
> ansible-playbook -i dev/inventory playbook.yml -t some-other-tasks --check
> ```

I am now really confused about your structure and playbooks. Could you clarify?

in0rdr: Sorry for the confusion. Let me expand on the current "solution" that "works for me" (without modifying the keepalived role):
```
$ tree
.
├── ansible.cfg
├── playbook.yml
├── deploy-haproxy.yml
├── deploy-keepalived.yml
├── dev
│   ├── group_vars
│   │   ├── haproxy.yml
│   │   └── keepalived.yml
│   ├── inventory
└── roles
    ├── ansible-role-haproxy
    └── keepalived
```

I seems like splitting the playbook into smaller ones makes the playbooks more user friendly as well.

The idea above mentioned for "Alternative 2" was to place the `redhat.yml` file (could be empty) in a directory/location where Ansible expects the file, because the error msg shows that it could not find the file ("No file was found when using first_found"). However, I could not find the correct location for the file.

We can close this PR anytime, just let me know. Or, if we want to investigate the issue further, let me know as well. I am fine with Ansible executing certain tasks `always`. I was just confused that Ansible assumed the keepalived role was the only role in the playbook and that it returned the strange error when executing custom tasks from the same playbook (that had nothing to do with the keepalived role). Instead of "always" executing the task, it "always" returned an error msg.

in0rdr: > ~what if you do -t always,some-other-tasks ?~ meh, let's ignore that comment. It shouldn't be that hard. Thanks ansible :s

That does not work, but I tried the following which works just fine to only execute the custom tasks:
```
ansible-playbook -i dev/inventory playbook.yml -t some-other-tasks --skip-tags=always --check
```

But as you mentioned, it is a bit cumbersome and not very straight forward to execute the playbook this way.

evrardjp: So for me, the structure is correct.
I would indeed create a new playbook for keepalived. But I would mostly include it in the haproxy playbook.
In OpenStack-Ansible [1], we simply ran the roles one after another, and if you want to not use one, you can just have a var you can flip in your variables or in the CLI. it's not really perfect. I thought in OSA we could also do it with tags though. 


[1]: https://github.com/openstack/openstack-ansible/blob/master/playbooks/haproxy-install.yml

evrardjp: I encourage you to not hide variables in a different file though, as it will be harder to maintain in the long run (where was this var set?). I am also afraid of the load variables location which could change, it happened in the past. So I would avoid skip as much as possible, and be very close to the standard.

evrardjp: Please note, I am not against a contribution, but we are here fixing something Ansible related, not something wrong with the code of the role. The role should indeed _always_ load those variables, to have an expected behaviour. The load variables shouldn't fail (why would it?), so no need to skip. It should be okay to use it with other plays. That's my 2cents.

If you're willing to dig deeper with it, I can help you. In any case, I propose we close this PR, as the role is not at fault, IMO, and create a new issue to tackle the fact the role can't behave nicely with only another tag.

in0rdr: that sounds very reasonable, I just created https://github.com/evrardjp/ansible-keepalived/issues/153 to sum up our findings. Many thanks for your help. Great role by the way!

PR #154, opened 2020-10-21
Ensure that keepalived is not masked

noonedeadpunk:
In cases when keepalived is currently masked, service
enable will not work. So service will
stay in masked state and won't start with handler either.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -202,6 +202,7 @@
   service:
     name: "{{ keepalived_service_name }}"
     enabled: "yes"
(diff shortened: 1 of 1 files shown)


evrardjp (approved):
I don't see a problem with that. Thanks for the fix!

evrardjp:
This was tagged and push to galaxy. Please bump your manifests to the minimum version of 3.13.1 to have this fix included!

Thanks @noonedeadpunk :)

noonedeadpunk:
Thanks for the fast merge:)

PR #155, opened 2020-10-28
fix systemtcl unit file refresh

spookiej:
Systemd unit files are modified and the daemon needs to be reloaded so that it can pick them up. In order for this to be restarted before the keepalived service, this needs to be the first task in the handlers file.

Current issue with the role will give you the following warning as the unit files have not been refreshed.

```
$ systemctl status keepalived
Warning: The unit file, source configuration file or drop-ins of keepalived.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -13,6 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- name: systemctl daemon-reload
+  systemd:
+    daemon_reload: true
+
 - name: restart keepalived
   service:
     name: "{{ keepalived_service_name }}"
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -13,11 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: systemctl daemon-reload
-  systemd:
-    daemon_reload: true
-
 - name: restart keepalived
   service:
     name: "{{ keepalived_service_name }}"
     state: restarted
+    daemon_reload: yes
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -226,6 +226,8 @@
   when:
     - ansible_service_mgr == 'systemd'
     - keepalived_systemd_overrides | bool
+  notify:
+    - systemctl daemon-reload
 
 - name: Apply keepalived override to restart service allways
   ini_file:
@@ -237,6 +239,8 @@
     - ansible_service_mgr == 'systemd'
     - keepalived_systemd_overrides | bool
     - keepalived_systemd_override_service_restart | bool
+  notify:
+    - systemctl daemon-reload
 
 - name: Remove keepalived overrides
   file:
@@ -245,3 +249,5 @@
   when:
     - ansible_service_mgr == 'systemd'
     - not (keepalived_systemd_overrides | bool)
+  notify:
+    - systemctl daemon-reload
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -227,7 +227,7 @@
     - ansible_service_mgr == 'systemd'
     - keepalived_systemd_overrides | bool
   notify:
-    - systemctl daemon-reload
+    - restart keepalived
 
 - name: Apply keepalived override to restart service allways
   ini_file:
@@ -240,7 +240,7 @@
     - keepalived_systemd_overrides | bool
     - keepalived_systemd_override_service_restart | bool
   notify:
-    - systemctl daemon-reload
+    - restart keepalived
 
 - name: Remove keepalived overrides
   file:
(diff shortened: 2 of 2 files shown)


handlers/main.yml
```diff
@@ -13,6 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- name: systemctl daemon-reload
+  systemd:
+    daemon_reload: true
```
evrardjp, line 18: did you try using daemon_reload in the service: restart keepalived handler?

evrardjp:
I am currently hitting https://github.com/ansible/ansible/pull/72337, preventing me to test this.

handlers/main.yml
spookiej: service module does not have the option for `daemon_reload` but probably should also restart the keepalived service too.

evrardjp: AFAIK, the service module is an action plugin that leverages to the right sysmgr module, which in this case, leverages the systemd module. At least it was :)

evrardjp: You're right the action plugin calls the legacy service module, and therefore doesn't expose daemon-reload. My bad.

evrardjp: What do you think of moving this block to the end of the main tasks https://github.com/evrardjp/ansible-keepalived/blob/master/tasks/main.yml#L201-L205 ? If we add a daemon_reload there, wouldn't it do what we need?

spookiej: Yes, the system module will use systemd if it is detected, but it will not do a daemon-reload.

Yeah sure, you probably also want to ensure that it isn't just enabled but it is also in the state `started`

evrardjp: I meant that:
```
service:
  name: "{{ keepalived_service_name }}"
  daemon_reload: "yes"
  state: "restarted"
```

Should work under systemd. Did you mean you tried it and it didn't work? In that case, we probably should file a bug at Ansible.

evrardjp: > Yeah sure, you probably also want to ensure that it isn't just enabled but it is also in the state `started`

That looks good to me.

spookiej: > I meant that:
> Should work under systemd. Did you mean you tried it and it didn't work? In that case, we probably should file a bug at Ansible.

Service module is a generic service module for all service managers and does not support `daemon_reload` as that is specific to systemctl.

Below are links to the supported options:
https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/service.py#L1579-L1592
https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/service.py#L20-L81

evrardjp: I disagree. The service is an action plugin which calls the service module if he doesn't know which plugin to load. In this case, the service action plugin uses the systemd module to do all the operations:  https://github.com/ansible/ansible/blob/fb09fd2a2301db485f5e15f77b59c31e7bc1645a/lib/ansible/plugins/action/service.py#L91

You can see that by running with a vvvv verbosity :)

spookiej: Yes whilst systemd supports daemon_reload the service module does not support issuing the command daemon-reload. There is no way to issue it using the service module.

evrardjp: What I am trying to tell you, is that service **does** work for that. See also a log here: http://paste.openstack.org/show/799532/ .

I explained the reason with my link above. Service is not only a module, it's also an action plugin. If you use the `service:` stanza in your playbooks, it will use the action plugin which will redirect to the appropriate module.

I am not sure how else I can explain that... Maybe you could believe me?

Anyway, my point was the following: We don't want to have more tasks around, especially not if it's just to daemon_reload. We should aim to _reduce_ the amount of tasks, not the other way around (as the more tasks you have, the longer the role takes to run). Here adding daemon_reload doesn't break readability, so I would suggest to do it in the service. Alternatively, if you don't believe me, just move to use the systemd module instead, which does make this whole operation a little bit faster by skipping the setup call.

spookiej: I fully understand what you are trying to say, but the service module itself does not support reloading of the module as a parameter. You would need to move to the systemd module to reload the daemon. 

I totally agree that you that we should not have more tasks to do the same thing.  At the end of the day I'm not fussed where you want to take the role, I mainly proposed this PR because the role had a bug where it modifies unit files and doesn't reload the daemon afterwards.

spookiej: I totally agree that there should not be any more tasks, I've update that task to reload the daemon, its really not clear in the documentation that it supports reloading of the daemon. I missed you previous link pointing to the `new_module_args` parameter.

does this look more suitable?

evrardjp (approved):
LGTM merging

evrardjp:
You can consume the latest version of the role from galaxy (3.13.2) which includes this fix.

spookiej:
Great! Thank you

PR #156, opened 2020-10-29
Restart keepalived on package change

noonedeadpunk:
In case keepalived_package_state is latest it is important to restart
service in case package has been upgraded

This also adds handlers flush once role is finished.

### DIFF
--- a/tasks/keepalived_install.yml
+++ b/tasks/keepalived_install.yml
@@ -92,6 +92,8 @@
     state: "{{ keepalived_package_state }}"
     update_cache: "{{ (ansible_pkg_mgr == 'apt') | ternary('yes', omit) }}"
     cache_valid_time: "{{ (ansible_pkg_mgr == 'apt') | ternary(cache_timeout, omit) }}"
+  notify:
+    - restart keepalived
   tags:
     - keepalived-apt-packages
 


tasks/main.yml
```diff
@@ -245,3 +245,6 @@
   when:
     - ansible_service_mgr == 'systemd'
     - not (keepalived_systemd_overrides | bool)
+
+- name: Flush handlers
```
evrardjp, line 249: I don't think we need the flush handlers, it will be automatically done at playbook/role's end.

Can you remove it?

noonedeadpunk: It will run at playbook end, not role end. Which means that you can't assume that keepalived will be restarted in case there're some another roles are to run by the play.

While it might be ok in some cases, this assumption is not correct imo. So, for example, in OSA case we restart keepalived only when we trigger handlers flush in haproxy role [1]

[1] https://zuul.opendev.org/t/openstack/build/f325c519da894d4b98e943e8ca521634/log/job-output.txt#16182

evrardjp: I believe the flush handlers should not be done in this role, as the role is "behaving" just fine. (This role is/was used outside OSA)
I am happy to add a meta: flush_handlers task in OSA in the playbook calling keepalived, after its role call. I think it's a better place.

If my memory serves well (I wrote this integration in OSA long ago!), I think the initial idea was to have one maintenance window, and therefore trigger all the handlers at one single point.

As you seem willing to restart keepalived before haproxy, I agree it can be changed by the introduction of a flush_handlers. This would be okay, as the downtime happening in case of an instance of keepalived restarting is close to 0 (ommitting misconfigurations). However, I would do that in OSA instead.

noonedeadpunk: Yeah, you;re right, we need to restart keepalived later as otherwise check scripts might go mad....

evrardjp (approved):
Makes sense. Merging.

evrardjp:
You can consume the latest version of the role from galaxy (3.13.2) which includes this fix.

Issue #126, opened 2020-01-14
Change role testing from Vagrant to molecule


evrardjp:
This will be simpler to test in Travis, and will be closer to upstream.

PR #157, opened 2020-10-30
Modernize role

evrardjp:
This is the future. Things have improved, let's remove workarounds that are all over the place.

Closes: #160 #126 

-----
[View rendered README.md](https://github.com/evrardjp/ansible-keepalived/blob/cleanup/README.md)


tasks/main.yml
```diff
@@ -243,5 +249,13 @@
     path: /etc/systemd/system/keepalived.service.d/override.conf
     state: absent
   when:
-    - ansible_service_mgr == 'systemd'
     - not (keepalived_systemd_overrides | bool)
+
+- name: Ensuring keepalived is enabled
```
noonedeadpunk, line 254: Why not to do all-in-one just in handler?

evrardjp: AFAIK, the service module is supposed to be either ensuring service state or ensuring enabled state. I had bugs in the past.
I also think it is probably better to always ensure the service is enabled, regardless of any change happening.

noonedeadpunk: I know what you're reffering to:) https://github.com/ansible/ansible/pull/68681

But it works wrong when you're trying to mask enabled service, while we're doing vice versa here. But ok, it's not something critical, just was wondering

evrardjp: oh well, I was thinking of another bug tbh, far older :) glad to see there are more :p

evrardjp:
Semaphore says it's broken, ofc, because it runs 14.04. Now upgraded to 18.04 to see what will happen.

tasks/main.yml
evrardjp: I have cleaned up the role of any "service" module/action plugin for now. It will be reverted when the bugs are fixed.

evrardjp:
Passes CI, let's merge this.

PR #159, opened 2020-11-02
Workaround https://github.com/ansible/ansible/pull/72337

evrardjp:
Without this, molecule testing and some other environments are
broken and can't install keepalived.

This should fix it until the fix above is backported to stable
Ansible branches.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -270,13 +270,19 @@
 - name: Check if keepalived is enabled
   shell: "systemctl daemon-reload && systemctl is-enabled {{ keepalived_service_name }}"
   register: isenabled
-  changed_when: isenabled.stdout == "disabled" and isenabled.rc != 0
-  failed_when: isenabled.stdout != "disabled" and isenabled.rc != 0
+  changed_when: isenabled.stdout == "disabled" or isenabled.stdout == "masked"
+  failed_when: (isenabled.stdout != "disabled" or isenabled.stdout != "masked") and isenabled.rc != 0
   when: ansible_service_mgr == 'systemd'
 
+- name: Unmask keepalived if necessary
+  command: "systemctl unmask {{ keepalived_service_name }}"
+  when:
+    - isenabled is changed
+    - isenabled.stdout == "masked"
+    - ansible_service_mgr == 'systemd'
+  
 - name: ensure keepalived is enabled
   command: "systemctl enable {{ keepalived_service_name }} --now"
-  changed_when: true
   when:
     - isenabled is changed
     - ansible_service_mgr == 'systemd'
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -244,6 +244,12 @@
     - ansible_service_mgr == 'systemd'
     - not (keepalived_systemd_overrides | bool)
 
+- name: Ensuring keepalived is enabled
+  service:
+    name: "{{ keepalived_service_name }}"
+    enabled: "yes"
+  when: ansible_service_mgr != 'systemd'
+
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -245,3 +243,34 @@
...
+# done in a single handler, tackling the daemon-reloads and the restart of the service,
+# plus a single task, handling the daemon-reload and ensuring the service is enabled
+# on the first install (the changes in systemd files would have to trigger the
+# handler to restart the service).
+- name: Check if keepalived is enabled
+  shell: "systemctl daemon-reload && systemctl is-enabled {{ keepalived_service_name }}"
+  register: isenabled
+  changed_when: isenabled.stdout == "disabled" and isenabled.rc != 0
+  failed_when: isenabled.stdout != "disabled" and isenabled.rc != 0
+  when: ansible_service_mgr == 'systemd'
```
noonedeadpunk, line 269: I think we won't have cases where service_mgr is not systemd. IIRC you've already dropped these conditionals before

tasks/main.yml
```diff
@@ -245,3 +243,34 @@
...
+
+- name: ensure keepalived is enabled
+  command: "systemctl enable {{ keepalived_service_name }} --now"
+  changed_when: true
+  when:
+    - isenabled is changed
+    - ansible_service_mgr == 'systemd'
```
noonedeadpunk, line 276: Ditto

tasks/main.yml
```diff
@@ -245,3 +243,34 @@
...
+# task to ensure keepalived is enabled, move it to an unconditional task here.
+# When the bug above is fixed with the service/systemd module, ensure everything is
+# done in a single handler, tackling the daemon-reloads and the restart of the service,
+# plus a single task, handling the daemon-reload and ensuring the service is enabled
+# on the first install (the changes in systemd files would have to trigger the
+# handler to restart the service).
+- name: Check if keepalived is enabled
+  shell: "systemctl daemon-reload && systemctl is-enabled {{ keepalived_service_name }}"
+  register: isenabled
+  changed_when: isenabled.stdout == "disabled" and isenabled.rc != 0
```
noonedeadpunk, line 267: What if service is masked?

tasks/main.yml
evrardjp: I have indeed dropped those in a different branch. My CI is/was testing ubuntu 14, so removing this breaks. I intend to refresh CI and remove those conditionals after this patch.

tasks/main.yml
evrardjp: good point, I need to fix this.

tasks/main.yml
evrardjp: TBH this has introduced a gap on old debian based systems where we don't ensure keepalived has been enabled.  I should fix that.

evrardjp: To be clear: I want this to be the latest release for 14.04, then I bump the versions to a new major to highlight the drop of support of upstart.

noonedeadpunk: Ah, ok then. It's just not supported for a year and a half, so I thought nobody cares about it:)

evrardjp: I do have some people who just contacted me this morning about where this role is heading :facepalm:

evrardjp: > Also I thought this bug occurs only with modern kernels (like 5.8) which 14.04 will never have?

So true. I am at risk of introducing bugs, for somethign that doesn't really matter for 14.04.

evrardjp: See also https://github.com/evrardjp/ansible-keepalived/issues/160 on why

PR #161, opened 2020-11-17
Add installation via snap

devster31:
I find snap a useful installation method for some binaries as it allows consistent, up-to-date versions across distribution and it's a bit easier to use than a container, hence this proposal.
I added this option to the role and a simple molecule scenario which should behave exactly like the default installation but use the snap binary.
I see two possible pending points:
* the inclusion of a separate role for snap setup and installation (e.g. `don_rumata.ansible_role_install_snap`) instead of including tasks in this role
* the addition of `community.general` to `meta/requirements.yml`; the role now uses `community.general.snap` and the molecule tests use `community.general.docker_network`.


tasks/main.yml
```diff
@@ -74,6 +133,19 @@
   notify:
     - restart keepalived
 
+- name: Configure keepalived service if snapd
```
evrardjp, line 136: Can you move that up in the snap section?

templates/keepalived.service.j2
```diff
@@ -0,0 +1,23 @@
+#jinja2: lstrip_blocks: True
```
evrardjp, line 1: can you rename this template keepalived-snap.service.j2 ? I think it makes it more obvious it's for snaps only.

tasks/main.yml
```diff
@@ -32,29 +32,88 @@
...
+    - name: Install keepalived from snap
+      community.general.snap:
+        name: keepalived
+        classic: true
+        channel: "{{ keepalived_snap_channel }}"
+
+    - name: Disable snap daemon
+      command: snap stop --disable keepalived
+      become: true
+      changed_when: false
```
evrardjp, line 104: I am reluctant to have a changed_when false where there is possibly a change. Is there no module that does this more appropriately, keeping idempotency and real status reporting?

evrardjp (changes requested):
I like the idea and implementation that contains its testing. Could you do the little changes I am talking about to improve the clarity? Thank you in advance.

evrardjp:
> I see two possible pending points:
> 
>     * the inclusion of a separate role for snap setup and installation (e.g. `don_rumata.ansible_role_install_snap`) instead of including tasks in this role
> 
>     * the addition of `community.general` to `meta/requirements.yml`; the role now uses `community.general.snap` and the molecule tests use `community.general.docker_network`.

To be honest, I would prefer _NOT_ include the snap setup in the role itself (it's fine in the molecule testing though). This is similar to the experience we have with selinux or many other necessary tools. What I mean by this, is that we should _document_ that for snap usage, the user should ensure all the snap requirements are setup, if necessary highlight how we've done it in our molecule testing. However, the role should not require other roles to be used.

I am fine to introduce a community general in meta requirements, as it seems it's the general tendency anyway. This role was refreshed recently, and we removed some old backwards compatibility, so i think it's a good time to do so now.

tasks/main.yml
devster31: https://github.com/evrardjp/ansible-keepalived/pull/161/commits/11378045cfc1e52b54371927f5458d181ffb487b should do it

templates/keepalived.service.j2
devster31: https://github.com/evrardjp/ansible-keepalived/pull/161/commits/60a5badcdc7c419692aa904e630aa5f44f2bfa0b should be it

tasks/main.yml
devster31: The module seems a bit old and simply runs commands remotely. I was unable to find something else.
I added an attempt to interact with the `snap` API with https://github.com/evrardjp/ansible-keepalived/pull/161/commits/ff2d15d9f9a8002be04fea57c7f75826ab67e0a5.

devster31:
I moved most of the `snap` setuo in the `converge.yml` playbook.
I also tried something different with https://github.com/evrardjp/ansible-keepalived/pull/161/commits/ff2d15d9f9a8002be04fea57c7f75826ab67e0a5 but it introduces dependencies on `community.general.json_query` and the `pip` package `jmespath` instead of `community.general.snap`.

tasks/main.yml
evrardjp: Ok for me!

tasks/main.yml
evrardjp: Sorry for any possible back and forth here. I supposed the module of snap is keeping idempotency and check mode.
Do you mean that the module isn't perfect and don't really do that?

If you're telling me the module is not perfect, I would prefer that we document _why_ we don't use it, so that we can re-evaluate when it's improved.

evrardjp:
I think ff2d15d is fine, but I want to really document why we do this before merging.

evrardjp:
Do you want me to add the "why" into this PR, so we can merge it?

tasks/main.yml
devster31: The module just installs and uninstalls snaps.
Some snaps, like this one, have "services" which are defined by the dev, which in the case of `keepalived` simply runs it with no particular frills. A service file is the better choice.
To my knowledge there is no module that stops/disables the snap daemon, which is required to use a custom service file.

devster31:
Sorry, missed the last message.
Sure, I commented above on the why I used the `command`, some snaps have "services" which are difficult to customize and not managed by a `systemd` file.
I haven't found an alternative to stop/disable these kind of daemons provided by snap besides my last attempt.

molecule/default/molecule.yml
```diff
@@ -10,10 +10,13 @@ driver:
 verifier:
   name: ansible
 provisioner:
+  env:
+    ANSIBLE_STDOUT_CALLBACK: yaml
```
evrardjp, line 14: While I appreciate the change, I believe it's not linked to the snap installation, so that should go into a separate commit / PR

molecule/snap/cleanup.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Prepare molecule instance
+  hosts: all[0]
+  gather_facts: no
+  tasks:
+  - name: Create a network with custom IPAM config
+    delegate_to: localhost
+    community.general.docker_network:
```
evrardjp, line 8: Would you mind using docker_network instead of `community.general.` ? 
While you're right in the fact it's the way forward, I believe it has to stay until I phase 2.9 out.

I also want to update everything in one go.

molecule/snap/prepare.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Prepare molecule instance
+  hosts: all[0]
+  tasks:
+  - name: Create a network with custom IPAM config
+    delegate_to: localhost
+    community.general.docker_network:
```
evrardjp, line 7: same comment as above.

evrardjp (changes requested):
A few changes.

Sorry for the delay.

PR #162, opened 2020-11-18
only reload if config has changed

gurubert:
a restart is not necessary


evrardjp (approved):
Completely fair!

evrardjp:
I am waiting for another open PR (which is close to merging) to merge before tagging them both and pushing the latest version on galaxy.

evrardjp:
This is included in 4.1.0.

PR #164, opened 2020-11-26
Add SMTP alert possibility

eifelmicha:
I like to have Email notifications when state changes. Didn't find yet the possibility in the keepalived template so i added it.

The "keepalived_global_defs" can already be used in current flavour like following example for defining email receivers:

```
    keepalived_global_defs:
      - notification_email_from {{ ansible_hostname }}
      - notification_email {{ mail@example.com }]
      - smtp_server smtp.example.com
```


templates/keepalived.conf.j2
```diff
@@ -204,4 +207,4 @@ virtual_server group {{ vserver_group.name }} {
   {% endfor %}
 }
 {% endfor %}
-{% endif %}
\ No newline at end of file
+{% endif %}
```
evrardjp, line 210: can you remove this change?

While this is technically okay (and better) to have an extra newline, doing so will trigger (for our existing users) a restart of keepalived (because the template has changed) on the next release, for no real good reason.

templates/keepalived.conf.j2
```diff
@@ -164,6 +164,9 @@ vrrp_instance {{ name }} {
   {% if instance.notify_stop is defined %}
   notify_stop "{{ instance.notify_stop }}"
   {% endif %}
+  {% if instance.smtp_alert is defined and instance.smtp_alert | bool %}
+  smtp_alert # Send email notification during state transition
```
evrardjp, line 168: Do we need that comment templated? Can't we just have it as a jinja comment instead? Is that comment necessary? It sounds obvious to me what smtp_alert does.

evrardjp (changes requested):
A few nits. Except those, I am fine with the change.

evrardjp:
(optional) This feature might be worth documenting in our README, what do you think?

templates/keepalived.conf.j2
eifelmicha: I will have a look. Seems my Editor automatically appends this.

templates/keepalived.conf.j2
eifelmicha: We don't need it, if you don't want it. I was having a looking at ```{% if instance.use_vmac is defined and instance.use_vmac | bool %}``` in the same file and there i found a comment.

evrardjp: Yeah that's more of a nit than anything. Don't you worry about that :)

evrardjp (approved):
Let's merge!

evrardjp:
Tagged and imported in galaxy :) 
Have a good day!

PR #166, opened 2020-12-22
Add config option "notify_master_rx_lower_pri"

eifelmicha:
Description taken from (https://xiaoz.co/2020/09/30/keepalived-and-gratuitous-arp/)

```
If a lower priority router has transitioned to master, there has presumably
been an intermittent communications break between the master and backup. It
appears that servers in an Amazon AWS environment can experience this.
The problem then occurs if a notify_master script is executed on the backup
that has just transitioned to master and the script executes something like
a `aws ec2 assign-private-ip-addresses` command, thereby removing the address
from the 'proper' master. Executing notify_master_rx_lower_pri notification
allows the 'proper' master to recover the secondary addresses.
```
I had some race conditions when calling a SOAP Service Endpoint to switch the FloatingIP to the masters. With this parameter this is now solved.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -130,6 +130,18 @@
   notify:
     - reload keepalived
 
+- name: Dropping the notification scripts for lower priority master case
+  copy:
+    src: "{{ item.value.src_notify_master_rx_lower_pri }}"
+    dest: "{{ item.value.notify_master_rx_lower_pri }}"
+    mode: "0755"
+  with_dict: "{{ keepalived_sync_groups }}"
+  when: item.value.src_notify_master_rx_lower_pri is defined
+  tags:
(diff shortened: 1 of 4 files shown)


evrardjp:
I am back from vacation. I will have a looksie now.

evrardjp (approved):
Good new feature.

Passes tests, and is documented. LGTM.

evrardjp:
The role's version 4.3.0 includes this commit and has been published on galaxy.

elcomtik:
@eifelmicha I had the same issues for months. When master backup communication failed, slave transitioned to master, but after recovery of communication high priority instance did not execute notify master script. Discovered that notify_master_rx_lower_pri solves issue.

Many thanks for implenmentation.

PR #167, opened 2021-02-08
Add a conditional check to configure virtual_ipaddress

mhmd3bdo:
Since `virtual_ipaddress` block isn’t required by keepalived configuration, it makes sense to be able to configure an 
instance without being required to configure `instance.vips`

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -110,11 +110,13 @@ vrrp_instance {{ name }} {
   {% endfor %}
   }
   {% endif %}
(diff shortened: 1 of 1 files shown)


evrardjp:
I am open to change that, but I haven't seen a keepalived in prod which doesn't use a VIP.
If you want to do that, we'll need to add a molecule test. This way we can cover the non-VIP and the VIP cases.

evrardjp:
As this change was done in your master branch, I am reluctant to rebase and push into your master branch.
I will do another PR to cover that case.

PR #168, opened 2021-03-01
Decrease TCP retries in case of VIP failover

noonedeadpunk:
In case of VIP failover some connections (like mysql) can stuck
in retrying to connect and detect a dead connection. We
should probably make this failover to be detected faster then the
default value as suggested in [1]

[1] https://access.redhat.com/solutions/726753

Related-Bug: https://bugs.launchpad.net/kolla-ansible/+bug/1917068

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -31,6 +31,15 @@ keepalived_instances: {}
 keepalived_sync_groups: {}
 keepalived_bind_on_non_local: False
 
+# In case of VIP failover some applications might stuck on retrying to send
+# data without updating ARP. This variable defines amount of retries kernel
+# will send before consider connection as dead. It will influence on VIP
+# failover speed directly. Default sysctl value is 15 which result in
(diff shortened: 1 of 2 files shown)


noonedeadpunk:
Um, where is CI?

evrardjp:
That's weird. It's supposed to be github actions.

evrardjp:
FYI that solution is unverified :)

I can see why we want to reduce it, but 3 is _really_ agressive. Should we move down to 8? 6? I think that below is very agressive. We might want to introduce a variable for the brave people out there :)

evrardjp:
This is on hold until #169 is confirmed working (fixing CI).

evrardjp:
See also: https://pracucci.com/linux-tcp-rto-min-max-and-tcp-retries2.html

evrardjp:
I think it's fair, and I will merge this.

Issue #170, opened 2021-03-02
Unknown keyword 'notify_master_rx_lower_pri'


elcomtik:
I just found out that notify_master_rx_lower_pri cannot be used in sync_group config section. It is ignored & it prints warnning at config load. More info in config reference https://manpages.debian.org/unstable/keepalived/keepalived.conf.5.en.html

Code which handles this should be removed until it gets support in keepalived eventually.
https://github.com/evrardjp/ansible-keepalived/blob/6a3fdd5704795d8485e8ce9301c348d57c69ea6c/templates/keepalived.conf.j2#L39-L41

evrardjp:
notify_master_rx_lower_pri is not defined by default, which means you shouldn't enter that loop unless you asked for it. Are you tellling me it gets templated anyway? AFAIK, keepalived doesn't have a config file validator, else we would validate the generated template before running.

elcomtik:
I defined it that way, because I had configured notify_master in sync_group section before. I looked only at template and thought it is supported to define notify_master_rx_lower_pri in sync_group section. However I found out recently, that it is not working like expected. It is partially my fault, but template shouldn't allow to generate config, which is not supporter. 

Keepalived handles this error silently, it starts with warning, but doesn't fail. I looked into man and foudn there is option --config-test, which may be useful for validation.

elcomtik:
Yeah, test arg may be recent feature. 

I couldn't test this use case, because I refactored my config and removed sync_group block, which I didn't need at all.  

We may add config-test to enhance role, maybe it saves someone from breaking things. I think, that this specific issue has simple solution, we should prevent from templating wrong config. notify_master_rx_lower_pri is supported only in instance block, which this role can template as well, this would at least prevent from misunderstanding of notify_master_rx_lower_pri usage.

evrardjp:
Oh yeah I see. Updated the template. If you could review that it works for you (in #171 ) that would be lovely.

elcomtik:
I just reviewed it. It works, many thanks for the quick action.

evrardjp:
I realised that patch was incomplete, so I will reopen this, sadly.

PR #171, opened 2021-03-03
Remove notify_master_rx_lower_pri from sync_group

evrardjp:
There is no such thing as notify_master_rx_lower_pri for
sync_groups.

This should fix it incorrect templates.

Closes: #170


elcomtik (approved):
I tested it, it works like expected 👍

evrardjp:
I created an issue on the validation, so we don't forget it.

PR #173, opened 2021-03-04
fix: Remove incorrect sync_group notify_master_rx_lower_pri

evrardjp:
The previous patch removed the incorrect templating, but didn't
remove the scripts and the examples in the documentation.

This should fix it.
Closes: #170

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -134,18 +134,6 @@
   notify:
     - reload keepalived
 
-- name: Dropping the notification scripts for lower priority master case
-  copy:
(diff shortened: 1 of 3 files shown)


elcomtik:
Oh no, I didn't realize that.

evrardjp:
My bad, I wasn't careful: too eager to fix to a point I didn't due my usual due diligence. :)

evrardjp:
Does that work for you, @elcomtik ?

elcomtik:
Good to go.

PR #176, opened 2021-05-25
Add support for accept/no_accept for instances.

naphta:
Coming to this role wanting to implement the keepalived configuration offered by [nginx plus](https://docs.nginx.com/nginx/admin-guide/high-availability/ha-keepalived-nodes/) and they have accept inside that configuration. Had a quick dig into the manpage and thought this should suffice.

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -93,6 +93,13 @@ vrrp_instance {{ name }} {
   {% if instance.dont_track_primary is defined and instance.dont_track_primary | bool %}
   dont_track_primary                               # Override VRRP RFC dont_track_primary default
   {% endif %}
+  {% if instance.accept is defined and instance.accept | bool %}
+  {% if instance.accept %}
+  accept
+  {% else %}
+  no_accept
+  {% endif %}
+  {% endif %}
   {% if instance.authentication_password is defined %}
   authentication {
     auth_type PASS


templates/keepalived.conf.j2
```diff
@@ -93,6 +93,13 @@ vrrp_instance {{ name }} {
   {% if instance.dont_track_primary is defined and instance.dont_track_primary | bool %}
   dont_track_primary                               # Override VRRP RFC dont_track_primary default
   {% endif %}
+  {% if instance.accept is defined and instance.accept | bool %}
```
evrardjp, line 96:
```suggestion
  {% if instance.accept is defined %}
```

templates/keepalived.conf.j2
```diff
@@ -93,6 +93,13 @@ vrrp_instance {{ name }} {
...
+  {% if instance.accept %}
```
evrardjp, line 97:
```suggestion
  {% if instance.accept | bool %}
```

templates/keepalived.conf.j2
evrardjp: This way if the user doesn't define accept (the default), there is no change in the config file.
And if someone writes instance.accept false, would write no_accept appropriately.

evrardjp (changes requested):
A few suggestions. Can you try those?

templates/keepalived.conf.j2
naphta: I've not tested the alternative (nor do I have an objection to it) but I'd have thought that my if which is checking if accept is defined and a boolean wouldn't pass in the case of `accept: `, thus not writing out `accept` or `no_accept` to the file.

PR #177, opened 2021-06-11
install dependencies required by keepalived scripts

mjourdan:
Hi,
Commands like "killall -0" are frequently used with keepalived. This PR aims to ease the installation of the packages providing those commands.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -32,9 +32,9 @@
   tags:
     - keepalived-config
 
-- name: install keepalived
+- name: install keepalived package(s)
   package:
-    name: "{{ keepalived_package_name }}"
+    name: "{{ [keepalived_package_name] + keepalived_scripts_packages }}"
     state: "{{ keepalived_package_state }}"
     update_cache: "{{ (ansible_pkg_mgr == 'apt') | ternary('yes', omit) }}"
     cache_valid_time: "{{ (ansible_pkg_mgr == 'apt') | ternary(cache_timeout, omit) }}"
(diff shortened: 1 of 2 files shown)


tasks/main.yml
```diff
@@ -44,6 +44,16 @@
   notify:
     - restart keepalived
 
+- name: install keepalived scripts dependencies
```
evrardjp, line 47: Can you add those to the module which installs the package, just above? No need to go twice through package manager.

Thank you!

mjourdan: Thanks for taking a look! The suggested change is done in b41410d

evrardjp: My pleasure! Not sure how I missed the notifications... I will have to dig into that.

evrardjp (changes requested):
If you squash your commits, we are good to go :)

mjourdan:
Done!

evrardjp:
I will dig into why it didn't pass tests.

evrardjp:
Sadly the fix will take more time than I expected, so I will not be able to fix this very soon. Sorry for the inconvenience. I will do my best to fix that in the next weeks.

evrardjp:
I will fix this when I will be back from vacay :)

evrardjp:
Back from vacation. Will work on this today or tomorrow.

evrardjp:
Hello. I hope you don't mind me pushing into your branch for updating the commit message and rebasing this commit.

mjourdan:
Hi @evrardjp, you did right! Thank you : )

PR #181, opened 2021-08-16
tags['always'] is bad

lorf:
Tasks with tags['always'] interfere with other roles, they are run when not asked to, which leads to this:
```
xxxx@xxxx:~/xxxx$ ansible-playbook -i inventory/ --tags etcd setup.yaml 

PLAY [elb] *******************************************************************************

TASK [evrardjp.keepalived : Gather variables for each operating system] **********************
fatal: [lb1]: FAILED! => {"msg": "No file was found when using first_found. Use errors='ignore' to allow this task to be skipped if no files are found"}
fatal: [lb2]: FAILED! => {"msg": "No file was found when using first_found. Use errors='ignore' to allow this task to be skipped if no files are found"}

PLAY RECAP ***********************************************************************************
lb1                        : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
lb2                        : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -22,7 +22,10 @@
     - "{{ ansible_distribution | lower }}.yml"
     - "{{ ansible_os_family | lower }}.yml"
   tags:
-    - always
+    - keepalived-apt-packages
+    - keepalived-config
+    - keepalived-install
+    - keepalived-packages
 
 - include: keepalived_selinux.yml
   when:


evrardjp:
I will have a look when I am back from vacation.

evrardjp:
Gathering variables should always be done.

Imagine if someone is using this role, and using `--tags keepalived-install` or `--tags keepalived-config` (for example to reload the config of keepalived after a change of settings). 
Without always, that task will not run, and the playbook/role will fail (especially if install is the tag selected)

I agree it could be solved by adding ALL the tags in the include_vars. Yet, people are _wrapping_ the role in some of their playbooks, adding tags to keepalived. I can't assume what they are doing there. However, I have seen people using both keepalived and haproxy in a single play. For those users, they have added extra tags for the keepalived role in their own playbook (e.g. 'keepalived') , and then run the play with --tags keepalived, to reconfigure keepalived.

That's where ```always``` is the right keyword: I can't assume what would people do in the role. The only way to _ensure_ that the role will behave properly in the latter case is to add ```always```.

What we need to solve is _why_ did that play fail for you. It should find the variables files. That's not the first time Ansible is bugged on finding paths btw.

Happy to hear an alternative solution, if you have one.

lorf:
In my case I have the central `setup.yaml` playbook with many roles included for various groups of hosts, `etcd` tag is set on the etcd role in the playbook and it is not related to keepalived in any way - etcd is installed on a different group of hosts. Running playbook with tag `etcd` skips keepalived hosts entirely, so facts, such as `ansible_distribution`, are not gathered for keepalived hosts and `Gather variables for each operating system` task from keepalived role fails.

So tasks tagged `always` interfere with other tags, which is why `always` is bad when used in roles. I think it's meant to be used in playbooks.

I don't have an alternative solution. But I don't see any negative effects, in theory and in practice, setting all available role tags on a task instead of the `always` tag.

evrardjp:
I will have to sync with different folks to make this happen.

noonedeadpunk:
There is a huge issue maintain that in a long run. As you as contributor as well as maintainer of the role should remember to add each new tag that is introduced. So I totally see how this will bring issues one day.

At the same time, you as role user still have control over what will run inside a playbook or how playbook is run have control over that already. So 2 ways I found in ~15 minutes over the issue:
1. Use `--skip-tags`. Example playbook:
```
- name: haproxy base config
  hosts: haproxy
  gather_facts: false
  serial: "{{ haproxy_serial | default('50%') }}"
  user: root
  roles:
    - role: "keepalived"
      when: haproxy_use_keepalived | bool
      tags:
        - keepalived
    - role: "haproxy_server"
      haproxy_service_configs: "{{ haproxy_default_services + haproxy_extra_services|default([]) }}"
      tags:
        - haproxy
```
And running it as ```ansible-playbook test.yml --tags haproxy --skip-tags keepalived``` will not include or run any task from keepalived

2. Use include_role

```
- name: haproxy base config
  hosts: haproxy
  gather_facts: false
  serial: "{{ haproxy_serial | default('50%') }}"
  user: root

  tasks:
    - include_role:
        name: "keepalived"
        apply:
          tags:
            - keepalived
      when: haproxy_use_keepalived | bool
      tags:
        - keepalived
    - include_role:
        name: "haproxy_server"
        apply:
          tags:
            - haproxy
      vars:
        haproxy_service_configs: "{{ haproxy_default_services + haproxy_extra_services|default([]) }}"
      tags:
        - haproxy
```

Then running simple as `ansible-playbook test.yml --tags haproxy` will omit any task inside keepalived role. The disadvantage of that though, is that you need to mention all tags from inside of the role in your playbook, otherwise these tags won't be respected.

evrardjp:
With these comments, I consider that there is a workaround available that might be easier to maintain in the long run. I will hence close this.

PR #184, opened 2021-10-03
add vmac_xmit_base support

howardZa:
Add support for the vmac_xmit_base option of keepalived (https://github.com/acassen/keepalived/blob/master/doc/NOTE_vrrp_vmac.txt)


evrardjp (approved):
I am fine with that.
I think it's untested from my side however.

PR #188, opened 2022-01-11
Ensure scripts are created before sync groups to prevent 'track scrip…

### LINKED ISSUE #187: (SG1) track script script_process not found, ignoring...
When using track_scripts in keepalived_sync_groups, the error "(SG1) track script script_process not found, ignoring..." is logged because the keepalived.conf.j2 template contains the vrrp_script section below the vrrp_sync_group section.

sigbjornaib:
Fixes #187
Ensure scripts are created before sync groups to prevent 'track script script_process not found' errors. 
Also enables use of the global_tracking option for sync groups.


evrardjp:
Hello.

I am a bit torn: If we bring this in, all the users of this role will have their keepalived restarted.
When it's about a change of file, that makes sense. When it's about whitespaces, it's a bit harder to justify.

However, I don't want to be slowing down progress here.
It feels it's _the right thing to do_. 

So, for the record, this patch will get merged, but we'll need to bump the version to map to that, so people are _aware that this will be intrusive_.

evrardjp:
If you could rebase those, that would be amazing, as it would pass tests ;)

evrardjp:
Is there any way I can help?

sigbjornaib:
Ok. I have rebased my branch with your current master.
I am not familiar with changing the commit message after the commit has been pushed to the repo.

evrardjp:
Ok, then allow me to work on this with you then.

sigbjornaib:
What do you need me to do? If you send me the steps I need to perform I will do my best to perform them.

evrardjp:
I have done the cleanup on my local branch.
Here is what I did:

`git rebase -i master` -> This removes the merge commit, put this PR on top

I wrote a commit to remove global_tracking, then another one to revert that removal.

I then rebased + squashed global_tracking remove on top of the first commit. Now my first commit and my last commits are clean: THe first one is focused on the move, the last one (The revert) is focused on the introduction of global tracking.

The I edited the second commit, and we're now good :)

evrardjp:
After double checking, I haven't found _anywhere_ the need to push vrrp_sync_groups after vrrp_script.

evrardjp:
I have updated your PR to be cleaner by force pushing into your branch. You can now see the result :)

evrardjp:
Updated role was pushed to galaxy just a few minutes ago.

PR #189, opened 2022-01-19
fix usage of deprecated include

zerwes:
fixes:
`[DEPRECATION WARNING]: "include" is deprecated, use include_tasks/import_tasks instead. This feature will be removed in version 2.16.`

### DIFF
--- a/tasks/keepalived_selinux.yml
+++ b/tasks/keepalived_selinux.yml
@@ -26,7 +26,8 @@
   when:
     - '"keepalived_ping" not in selinux_modules.stdout'
 
-- include: keepalived_selinux_compile.yml
+- include_tasks:
+    file: keepalived_selinux_compile.yml
   when:
     - selinux_policy_name not in selinux_modules.stdout
   with_items: "{{ keepalived_selinux_compile_rules }}"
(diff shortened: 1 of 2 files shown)


evrardjp:
Hello.
You're 100% right, the code itself isn't the cause of a drop of 2.9. 
Our testing framework is. I took the decision to move away from 2.9 because we would accrue tech debt, rather than reducing it.

please rebase your code, and the testing should work now.

evrardjp:
@zerwes If there is an issue in testing with debian11, would you mind filing an issue for it, with the logs? We'll try to figure it out. It's probably a molecule thing again.

evrardjp:
Commit message can be improved, but I won't be fussy about it. It's explaining the "what and why" enough.

zerwes:
Thanks @evrardjp for the fixed test. All [molecule tests](https://github.com/Rosa-Luxemburgstiftung-Berlin/ansible-keepalived/actions/runs/1725216003) pass after rebasing the PR branch.

evrardjp:
I will bump the role on galaxy when another PR about template is merged.

evrardjp:
Updated role was pushed to galaxy just a few minutes ago.

PR #195, opened 2022-04-01
Ensure instance.vips_excluded is not empty

tridde:


evrardjp:
Can you rebase ?

evrardjp:
@tridde  if you could rebase that would be great :)

PR #196, opened 2022-05-02
Add ansible_managed output to keepalived.conf

senuphtyz:
Output is by default enabled but can be disabled
using variable keepalived_show_ansible_managed set to false.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Wheter to add {{ ansible_managed }} information on top of keepalived.conf
+keepalived_show_ansible_managed: false
+
 # If running keepalived with SELinux, you could need to compile your
 # rules. Please override this list with path to files to compile.
 keepalived_selinux_compile_rules:
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -14,6 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 -#}
+{% if keepalived_show_ansible_managed|bool %}
+#
+# {{ ansible_managed }}
+#
+{% endif %}
 
 {% if keepalived_global_defs is defined %}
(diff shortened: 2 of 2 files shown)


templates/keepalived.conf.j2
```diff
@@ -14,6 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 -#}
+{% if keepalived_show_ansible_managed %}
```
evrardjp, line 17: could you add `| bool` ?

evrardjp (changes requested):

templates/keepalived.conf.j2
senuphtyz: Sounds a good idea .. changed.

defaults/main.yml
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Wheter to add {{ ansible_managed }} information on top of keepalive.conf
```
evrardjp, line 16: I am sorry I missed this in my first review:
Can you:
1) Update the comment to mention keepalived.conf instead of keepalive.conf
2) Make it `false` by default?

The reason to make it false would be that the existing users who would update the role would not have a "changed" behaviour, which would restart keepalived, for something they didn't opt-in.

Sorry for the inconvenience.

evrardjp (changes requested):

defaults/main.yml
senuphtyz: I fully understand your argument and i absoluty agree with.

templates/keepalived.conf.j2
evrardjp: Thanks a lot!

defaults/main.yml
evrardjp: Thanks for your understanding :)

evrardjp (approved):

evrardjp:
Thanks, great contribution!

Issue #197, opened 2022-05-02
Linting fails since ansible lint is updated


evrardjp:
We now have the problem that we can't use the "short" name for tasks.

See also the lint failures: `fqcn-builtins Use FQCN for builtin actions.`

evrardjp:
See also:

- https://github.com/ansible/ansible-lint/pull/1614
- https://github.com/ansible/ansible-lint/pull/1908
- https://github.com/ansible/ansible-lint/issues/2050

zerwes:
yes, I like those breaking changes in ansible too :-/

but with the simple fix from the PR +199
```diff
+ All checks have passed
+ 3 successful checks
```

PR #199, opened 2022-05-02
FIX "fqcn-builtins Use FQCN for builtin actions." on ansible lint run

zerwes:
added fqcn-builtins to ansible-lint skip_list


zerwes:
fixes #197

PR #202, opened 2022-05-30
Make "auth_type" variable optional with default value

maxkochubey:
In case if user needs to change the `auth_type` variable value to `AH`.

### DIFF
--- a/templates/keepalived.conf.j2
+++ b/templates/keepalived.conf.j2
@@ -110,8 +110,8 @@ vrrp_instance {{ name }} {
 
   {% if instance.authentication_password is defined %}
   authentication {
-    auth_type PASS
-    auth_pass {{instance.authentication_password}}
+    auth_type {{ instance.authentication_type | default('PASS') }}
+    auth_pass {{ instance.authentication_password }}
   }
   {% endif %}
   {% if instance.unicast_src_ip is defined and instance.unicast_peers|length > 0 %}


evrardjp:
Do you want to include tests, to ensure that this is working at all times?
Or are you okay with the feature as of now?

evrardjp:
the molecule part isn't the hardest I would say.
What matters here would be to know how to ensure this works.

I don't think we should test that keepalived delivers their code right (there must be tests in keepalived for that), but a test to ensure that the configuration is _as expected_ .

If you tell me that it's very close to the rest of the config, and that we shouldn't ensure that, then I guess we can merge as is.

maxkochubey:
@evrardjp, yes, it works - here is the working example used in my lab.

Playbook:
```yaml
- hosts: keepalived
  become: yes

  pre_tasks:
    - name: Create a common keepalived password
      tags: [ 'keepalived' ]
      set_fact:
        keepalived_password: "{{ (keepalived_vip|string+keepalived_vr_id|string)|hash('md5')|truncate(8, True, '') }}"
      delegate_to: "{{ groups['keepalived'][0] }}"

    - name: Adding AH and VRRP protocols to the firewall whitelist
      tags: [ 'fwrules' ]
      ansible.posix.firewalld:
        rich_rule: "{{ item }}"
        zone: public
        permanent: yes
        immediate: yes
        state: enabled
      loop:
        - 'rule protocol value="ah" accept'
        - 'rule protocol value="vrrp" accept'

  roles:
    - role: ansible-keepalived
      tags: [ 'keepalived' ]
      vars:
        keepalived_instances:
          floating_ip:
            interface: "{{ keepalived_nic }}"
            state: "{{ (groups['keepalived'].index(inventory_hostname) == 0) | ternary('MASTER','BACKUP') }}"
            virtual_router_id: "{{ keepalived_vr_id }}"
            priority: "{{ (groups['keepalived']|length - groups['keepalived'].index(inventory_hostname)) * 250 // (groups['keepalived']|length) }}"
            authentication_type: "AH"
            authentication_password: "{{ hostvars[groups['keepalived'][0]]['keepalived_password'] }}"
            vips:
              - "{{ keepalived_vip }} dev {{ keepalived_nic }} label {{ keepalived_nic }}:vip"
```

So the master config looks like the following after deployment:
```
vrrp_instance floating_ip {
  interface eth0
  state MASTER
  virtual_router_id 42
  priority 250

  authentication {
    auth_type AH
    auth_pass 8e22f47e
  }

  virtual_ipaddress {
    10.10.10.10/32 dev eth0 label eth0:vip
  }
}
```

evrardjp:
Ok, I propose we roll this out without extra tests, and we'll see with usage :)

evrardjp:
I will try to fix the CI this weekend by pinning the version of ansible lint, then propose extra patches. Thanks for your patience.

evrardjp:
"This weekend" was 27 days ago. I am sorry. I will still work on that soon (TM).

maxkochubey:
Don't worry and take your time, thanks :)

evrardjp (approved):
Let's merge this. If there is a need for extra testing in the future, we'll see.

PR #211, opened 2022-07-13
Remove deprecation warning "Use 'ansible.utils.ipaddr' module instead"

zerwes:


evrardjp (approved):
Good patch.

Need to fix the occasional issue with linting now...

PR #213, opened 2022-07-13
use reload-or-restart for service reload to ensure a running service

zerwes:
git cherry-pick e14b518 from pr #198 (as the other commits in the pr had some implications)

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -25,7 +25,7 @@
 #    name: "{{ keepalived_service_name }}"
 #    state: restarted
 - name: reload keepalived
-  ansible.builtin.command: "systemctl reload {{ keepalived_service_name }}"
+  ansible.builtin.command: "systemctl reload-or-restart {{ keepalived_service_name }}"
   tags:
     - skip_ansible_lint
 #- name: reload keepalived


zerwes:
regarding the failing centos8 check:
```diff
- fatal: [keepalived-centos8]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'baseos': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried", "rc": 1, "results": []}
```
is centos 8 not dead and archived since end of 2021?

evrardjp:
Fair point, but I didn't add any alternatives yet...

evrardjp:
Could you watch this (alternative implementation) in the draft PR I created yesterday? https://github.com/evrardjp/ansible-keepalived/pull/212/files

It is intending to deliver the same feature set, tested through molecule, with very simple ansible code.

zerwes:
Hello @evrardjp : sorry, I did not realized the pr #212 ... looks promising ... good luck and success with the approach
I just had the situation, where the reload-or-restart (implemented in the branch we use at the moment) just saved me again, so I was thinking to push just this single commit without the other implications in the pr #198
btw: [here](https://github.com/zerwes/ansible-role-checkmkagent/commit/f37c87952ca44dfb6d65627eebce91b4c5d23ab4) I have implemented the service state without handlers at all using a [registered desired service state](https://github.com/zerwes/ansible-ensure-service-running#register-the-desired-state-of-the-service-and-apply-it) ... the test is done in molecule [side_effect](https://github.com/zerwes/ansible-role-checkmkagent/blob/main/molecule/default/side_effect.yml), but the test is quite specific.
So I am closing this for now

evrardjp:
I think the approach of the registering the desired state in a fact is more elegant. 
Yet, the problem I see is that when you multiply the amount of tasks, the set_fact will contain a lot of when: 
"x is changed or y is changed or z is changed or a is changed or b is changed" . Which will start to be a reimplementation of the handlers.

IMO, I prefer the explicit state, but I think contributors might find it less "obvious", so I am willing to stick with notify, as I found a workaround that works on my env. I indeed use the side_effect to check if reload is used, and it seems it's the case.

Issue #192, opened 2022-01-20
issue in testing with debian 11 aka bullseye


zerwes:
@zerwes If there is an issue in testing with debian11, would you mind filing an issue for it, with the logs? We'll try to figure it out. It's probably a molecule thing again.

_Originally posted by @evrardjp in https://github.com/evrardjp/ansible-keepalived/issues/189#issuecomment-1017260054_

zerwes:
while running `molecule test` on debian 11 with
```
$ molecule --version
molecule 3.5.2 using python 3.9 
    ansible:2.12.1
    delegated:3.5.2 from molecule
    docker:1.1.0 from molecule_docker requiring collections: community.docker>=1.9.1
```
I get in the converge step
`INFO     Running default > converge`
a error related to systemd not beeing available in the docker image
```
TASK [ansible-keepalived : Check if keepalived is enabled] *********************
fatal: [keepalived-bionic]: FAILED! => {"changed": false, "cmd": "systemctl daemon-reload && systemctl is-enabled keepalived.service", "delta": "0:00:00.004321", "end": "2022-01-20 19:34:58.212965", "failed_when_result": true, "msg": "non-zero return code", "rc": 1, "start": "2022-01-20 19:34:58.208644", "stderr": "Failed to connect to bus: No such file or directory", "stderr_lines": ["Failed to connect to bus: No such file or directory"], "stdout": "", "stdout_lines": []}
```
(seems to fail on every image i.e.:
```
TASK [ansible-keepalived : Check if keepalived is enabled] *********************
fatal: [keepalived-centos7]: FAILED! => {"changed": false, "cmd": "systemctl daemon-reload && systemctl is-enabled keepalived", "delta": "0:00:00.469191", "end": "2022-01-20 20:55:46.937047", "failed_when_result": true, "msg": "non-zero return code", "rc": 1, "start": "2022-01-20 20:55:46.467856", "stderr": "Failed to get D-Bus connection: No such file or directory", "stderr_lines": ["Failed to get D-Bus connection: No such file or directory"], "stdout": "", "stdout_lines": []}
```
I thing this is related to the geerlingguy images ... (I can recall that in another context I had some issues with them too), but docker and molecule are not my area of expertise, so this might be only half of the truth ...

[...]

PR #215, opened 2022-07-14
Update default scenario OSes

evrardjp:
CentOS 7 is outdated and sometimes fail due to flakey mirrors.
Need to update the OS list to something more recent.

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -25,7 +25,6 @@ galaxy_info:
   platforms:
   - name: "EL"
     versions:
-    - "7"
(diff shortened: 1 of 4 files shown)


zerwes:
Regarding changing the OS to debian 11: +1 - glad to see that.
But before it drives you nuts, be aware of #192 and https://github.com/geerlingguy/docker-debian11-ansible/issues/4

evrardjp:
Seems like it's passing testing, so i guess I am good to go :)

Issue #186, opened 2021-10-11
Improve visibility for ansible-keepalived on galaxy


evrardjp:
not that it matters much... but sometimes i have trouble myself to find this role on galaxy.

Would be good if we work on optimising this: Probably adapt the tags, and follow what's necessary to stay on top.

We have tons of downloads, it makes sense to not appear on page over 9000.

evrardjp:
I will move this to a discussion.

PR #218, opened 2022-07-15
Add more meta tags to improve visibility on galaxy

evrardjp:
This should get us to the top of the page 2.
Next step is to include this role in a collection to reach page 1.


evrardjp:
See also: #186 and discussion in this repo.

PR #226, opened 2022-07-19
Use ansible_facts instead of global scoped variables

evrardjp:


noonedeadpunk (commented):
LGTM

tasks/main.yml
```diff
@@ -29,8 +34,8 @@
     file: keepalived_selinux.yml
   when:
     - keepalived_selinux_compile_rules | length > 0
-    - ansible_selinux.status is defined
-    - ansible_selinux.status == "enabled"
+    - ansible_facts['selinux']['status'] is defined
```
noonedeadpunk, line 37: Oh, well, I'd probably did smth like

- "'selinux' in ansible_facts"

evrardjp: it is always in ansible_facts, even without selinux. Status should be defined, but I kept it for safety.

noonedeadpunk: Huh, I thought that for Debian it should not be there...

PR #232, opened 2022-07-29
Add support for checkmode in tasks/main.yml

sd-hardy:
I'm not able to run the playbook in check mode due to the "Output keepalived version" task. ansible.builtin.command doesn't run in check mode because it assumes changes are being made. No changes are being made in this case, so I've added 'check_mode: no'.

Docs for the check_mode option are here: https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html#enforcing-or-preventing-check-mode-on-tasks

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -60,6 +60,7 @@
   ansible.builtin.command: keepalived --version
   register: _keepalived_version
   changed_when: false
+  check_mode: no
   tags:
     - keepalived-config
 


evrardjp:
Great addition -- I am surprised molecule didn't catch that. 
Do you think there is a way to catch those? Should we add a test?

evrardjp:
@sd-hardy  If you want to ensure check mode always behave as you expect, please don't hesitate to submit a follow up PR :)

Thanks for that patch @sd-hardy .

sd-hardy:
I'm not too familiar with Molecule, but I did find there is a way to add check mode to the testing sequence: https://github.com/ansible-community/molecule/issues/128

PR #240, opened 2022-09-19
Fixes conditional logic and other issues with dropping scripts

busterswt:
This patch addresses a few issues with recent changes, including:

- The 'validate' template directive fails due to check scripts not yet being dropped
- The scripts not being dropped due to the conditional not being met
- keepalived_scripts being changed to keepalived_sync_groups (inadvertently?)

NOTE: This patch only addresses the tracking scripts, but may need to be modified to address similar issues with other notification scripts.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -125,17 +125,6 @@
   tags:
     - keepalived-install
 
-- name: Configure keepalived
-  ansible.builtin.template:
-    src: keepalived.conf.j2
-    dest: "{{ keepalived_config_file_path }}"
-    mode: "0640"
-    validate: "{{ keepalived_config_testable | ternary('keepalived --config-test -f %s', omit) }}"
-  notify:
-    - reload keepalived
-  tags:
-    - keepalived-config
-
 - name: Check that daemon options file exists
   ansible.builtin.stat:
     path: "{{ keepalived_daemon_options_file_path }}"
@@ -161,8 +150,8 @@
     src: "{{ item.value.src_check_script }}"
     dest: "{{ item.value.dest_check_script | default(item.value.check_script) }}"
     mode: "0755"
-  loop: "{{ keepalived_sync_groups | dict2items | default('{}') }}"
-  when: item.value.src_check_script is defined
+  loop: "{{ keepalived_scripts | dict2items | default('{}') }}"
+  when: "'src_check_script' in item.value"
   notify:
     - reload keepalived
   tags:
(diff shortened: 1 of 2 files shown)


noonedeadpunk (changes requested):
Same issue is present on L177, L189,L201,L213,L225,L237 in original code or L166,L178,L190,L202,L214,L226 with patch applied

noonedeadpunk:
Interesting though, that keepalived_scripts never were a thing in defaults. The only reference is in tests:
https://github.com/evrardjp/ansible-keepalived/blob/ebbeac6c27de5cccd6d9c9c3992e5a2acd15a86e/tests/keepalived_haproxy_backup_example.yml#L44-L59

So I assume that change to keepalived_sync_groups is valid, though it was performed without proper description or squashed together with conditional change.

But I'd say it's rather OSA that needs to adopt to usage of keepalived_sync_groups

noonedeadpunk:
Ok, disregard, you're right, I guess it was a typo after looking at the tempalte.

noonedeadpunk (dismissed):
LGTM

evrardjp:
> Ok, disregard, you're right, I guess it was a typo after looking at the tempalte.

For the record, yes, you don't need 'external' scripts to be copied to have _checks_ working. The examples are indeed using binaries that are generally shipped on unixes.

tasks/main.yml
```diff
@@ -161,8 +150,8 @@
     src: "{{ item.value.src_check_script }}"
     dest: "{{ item.value.dest_check_script | default(item.value.check_script) }}"
     mode: "0755"
-  loop: "{{ keepalived_sync_groups | dict2items | default('{}') }}"
-  when: item.value.src_check_script is defined
+  loop: "{{ keepalived_scripts | dict2items | default('{}') }}"
```
evrardjp, line 153: keepalived_scripts | default({}) | dict2items.

Without this, keepalived_scripts is by default undefined. You can probably add keepalived_scripts: {} in defaults if you find this more clean...

evrardjp (changes requested):
Need to change the loop.
Without it, it will fail with the following:
```
{"msg": "Unable to look up a name or access an attribute in template string ({{ keepalived_scripts | dict2items | default('{}') }}).
Make sure your variable name does not contain invalid characters like '-': dict2items requires a dictionary, got <class 'ansible.template.AnsibleUndefined'> instead."}
 ```

tasks/main.yml
noonedeadpunk: Or we can add `keepalived_scripts` to defaults/main.yml with some example so that it was more obvious that this variable is available for use?

evrardjp: Kinda close to what I have said a bit earlier? :)
The documentation is not in defaults, but simply in README. I don't mind a duplication here.

evrardjp: Side note: I indeed prefer to have an example in defaults/, and refer to it in README. :)

busterswt: i will update the readme to provide an example (that's already provided in /tests)

busterswt:
@evrardjp Let me know if those changes are what you had in mind

evrardjp (approved):
This role isn't great, but you made it better @busterswt . Thanks for that!

https://giphy.com/gifs/chuck-norris-thumbs-up-BIuuwHRNKs15C

PR #256, opened 2022-12-15
Commenting out unused variables for shellcheck compliance.

crazzy:
Found this role being used in a project I'm collaborating on, where I want to run shellcheck on every single shellscript, so figured I'd patch this one and submit a PR.

### DIFF
--- a/files/keepalived_notifications.sh
+++ b/files/keepalived_notifications.sh
@@ -13,11 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-TYPE=$1
-NAME=$2
+#TYPE=$1
+#NAME=$2
 NOW=`date "+%Y-%m-%d %H:%M:%S"`
 NEWSTATE=$3
-OLDSTATE=$(cat /var/run/keepalived.state)
+#OLDSTATE=$(cat /var/run/keepalived.state)
 
 echo "$NEWSTATE" > /var/run/keepalived.state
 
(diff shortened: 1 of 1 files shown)


files/keepalived_notifications.sh
```diff
@@ -13,11 +13,11 @@
...
 # limitations under the License.
 
-TYPE=$1
-NAME=$2
+#TYPE=$1
+#NAME=$2
 NOW=`date "+%Y-%m-%d %H:%M:%S"`
 NEWSTATE=$3
-OLDSTATE=$(cat /var/run/keepalived.state)
+#OLDSTATE=$(cat /var/run/keepalived.state)
```
evrardjp, line 20: You don't need to add this again. I am not sure why it's still there. This was just an example anyway, so feel free to remove the line altogether.

evrardjp (dismissed):
Thanks for the review. Good catch.

files/keepalived_notifications.sh
crazzy: Absolutely, I've removed that line now.

evrardjp:
Sorry for the late answer.

You can see the merge wasn't done due to failure in CI.

Here is the PR of the CI change. https://github.com/evrardjp/ansible-keepalived/pull/259 . 

When this is merged, I will rebase your code and run it again. If all goes well, it will be merged this week.

evrardjp:
I just noticed that shellcheck isn't included in the tests, would you be okay to add it? Probably need to add somethign on the runners (.github/workflow/) and in the tox.ini (the shellcheck invocation)

PR #259, opened 2022-12-29
Fix ci

evrardjp:


evrardjp:
Of course the docker images broke again. :/

evrardjp:
This is only a partial fix. 

Moving to another fix will take long time, so I will force merge this.

PR #261, opened 2022-12-29
Cleanup Examples

evrardjp:
The examples were a bit too complex to master for many, so this
should clarify the documentation.


evrardjp (commented):
Seems like my renaming got taken as a new file and deletion of previous file. All good, I will just update the copyright date.

evrardjp (commented):
Looking better now

PR #264, opened 2023-01-12
Support absolute paths for SELinux modules

cosandr:
Hi,

I like to keep forks of roles to a minimum and often refer to files (like your `keepalived_selinux_compile_rules` list) by absolute paths, usually based on `inventory_dir` or similar. This isn't supported currently as it assumes each item in that list is a filename. This PR adds a new var for the basename which is used in most places. It has no effect on the default values and shouldn't be a breaking change for existing users as I expect it currently fails with relative paths like it does with absolute ones.

I've also changed the check used for installing SELinux packages to check against the first element in `keepalived_selinux_compile_rules` instead of hardcoding `keepalived_ping`, which might be removed by a user overriding the list and thus missing the requirements for compiling modules.

### DIFF
--- a/tasks/keepalived_selinux.yml
+++ b/tasks/keepalived_selinux.yml
@@ -24,7 +24,7 @@
     name: "{{ keepalived_selinux_packages }}"
     state: present
   when:
-    - '"keepalived_ping" not in selinux_modules.stdout'
+    - keepalived_selinux_compile_rules[0] not in selinux_modules.stdout
 
 - name: Compile SELinux rules
   ansible.builtin.include_tasks:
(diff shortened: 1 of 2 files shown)


evrardjp:
I sadly don't have the time to check this in the next weeks. 

As selinux coverage was never big, I can't ensure test this myself. Can you confirm it's been tested okay ?

When I will be available, I'll be happy to review and merge as I don't have anything against the idea.

Or, in other words: Looks good at first sight, but need to spend more time on it.

cosandr:
No worries, I'm not in a rush.

I've tried it by removing all modules it installed and it worked with the default values as well as my own custom policy which is using an absolute path.

evrardjp (approved):
That's better indeed.

PR #269, opened 2023-02-20
Update keepalived-install-example.yml

Eamixs:


evrardjp:
It looks better indeed. Do you think we should clarify the VRRP priorities? Many ppl get them wrong.

PR #274, opened 2023-03-14
Compability with ansible-lint, yamllint and ShellCheck

crazzy:
Since I use this role as a submodule in a CI environment which enforces strict compability with these tools I've fixed all errors and warnings in both yamllint, ansible-lint and ShellCheck.


evrardjp:
I have left this rot for too long. Closing it now. Thanks for the commit @crazzy .

PR #275, opened 2023-03-24
Support virtualserver/inhibit_on_failure

dhoffend:
We need to set the flag `inhibit_on_failure` in our environment.


evrardjp (approved):
Looks okay to me

PR #276, opened 2023-03-27
enable track_interface config per sync_group

zerwes:


evrardjp:
Hello @zerwes! Welcome back :) 

Thanks for your contribution. 

As its not the first time you are contributing, would you be inclined to become maintainer of the role?

evrardjp (approved):
This makes sense to me.

zerwes:
> As its not the first time you are contributing, would you be inclined to become maintainer of the role?

Hello @evrardjp

Thank you very much for the offer and the trust that comes with it.
I'm quite busy with family, job, my own projects, garden etc., so my resources might be limited. And my dev focus is primarily on debian etc. But if I can be of help, I am willing to invest some time here, as I consider the role quite important and useful.
So if I can help here, juts let me know how ...
Greetings

PR #292, opened 2023-08-24
Add useful parameters to vserver and checks

matchaxnb:
Hello @evrardjp, Jean-Philippe,

I have added 2 parameters to your module:

`vserver.quorum`: allows to define the quorum weight needed to operate properly. If this quorum is not met, the sorry_server will respond instead. This allows for split brain avoidance mechanisms with a witness tie-breaker for example.

`*_check.delay_loop`: this parameter defines the base frequency with which health checks are done. `delay_before_retry` is an override for this parameter only in the case of failures. By setting `delay_loop` to an integer, the healthcheck will be run every that-much seconds.

I hope you like it, find it up to your standards, and merge it.

### DIFF
--- a/templates/keepalived-checks.j2
+++ b/templates/keepalived-checks.j2
@@ -12,6 +12,9 @@
       {% if tcp_check.retry is defined and tcp_check.retry | int %}
       retry {{ tcp_check.retry }}
       {% endif %}
+      {% if tcp_check.delay_loop is defined and tcp_check.delay_loop | int %}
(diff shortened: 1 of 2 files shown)


evrardjp (changes requested):
It looks good overall. I would prefer if you would amend your commit with the port though, and push again on this branch.

evrardjp:
I don't want to block this too much, it's detailed if it's all merged into a single MR.

evrardjp (approved):

PR #293, opened 2025-01-24
Feature: Install Keepalived from Source

Salvoxia:
I was trying to install and configure keepalived with this role. However, I ran into issues because my package manager only serves a pretty old version of keepalived (2.2.7), which caused troubles e.g. during configuration verification.
So I extended the role to allow for compiling and installing keepalived from source.

The role now supports the following scenarios:
  - Compile and install Keepalived from a Git repository with a specific source tag
  - Uninstall Keepalived after installed from source
  - Switch from installation by a package manager to installation from source
  - Switch from installation from source to installation by a package manager
  - Switch from one installed version from source to a different version

The follwing role variables control installation from source:
```yaml
# Flag indicating whether to compile and install Keepalived from source instead of a package manager
keepalived_install_from_source: false
# Git source repository to use when installing Keepalived from source
keepalived_source_repository: https://github.com/acassen/keepalived.git
# The Git tag to compile when installing Keepalived from source
keepalived_source_tag: v2.3.2
# Configure options
keepalived_source_configure_options:
    - --includedir=${prefix}/include
    - --mandir=${prefix}/share/man
    - --infodir=${prefix}/share/info
    - --sysconfdir=/etc
    - --localstatedir=/var
    - --disable-option-checking
    - --disable-silent-rules
    - --runstatedir=/run
    - --disable-maintainer-mode
    - --disable-dependency-tracking
    - --enable-snmp
    - --enable-sha1
    - --enable-snmp-rfcv2
    - --enable-snmp-rfcv3
    - --enable-dbus
    - --enable-json
    - --enable-bfd
    - --enable-regex
```
The installation prefix is generated dynamically using `keepalived_source_tag`.

In addition, the following variable control the build dependencies on a distribution level:
```yaml
keepalived_build_dep_packages:
  - git
  - autoconf
  - curl 
  - gcc 
  - libssl-dev 
  - libnl-3-dev 
  - libnl-genl-3-dev 
  - libsnmp-dev
  - libsystemd-dev
  - libmnl-dev
  - libipset-dev
  - libnfnetlink-dev
  - libnl-route-3-dev
```

In addition to that, I also revamped the molecule tests. The ones that already existed are working now, and I also added a couple of my own for testing source installation.
The tests now [run successfully](https://github.com/Salvoxia/ansible-keepalived/actions/workflows/molecule.yml) on the following platforms:
  - CentOS 8
  - Debian 10
  - Debian 11
  - Debian 12
  - Ubuntu 22.04
  - Ubuntu 24.04

RockyLinux 9 also works, but I had to disable the automated tests [due to this issue](https://github.com/geerlingguy/docker-rockylinux9-ansible/issues/6).

Please let me know if you don't have any interest in merging these features. Then I'll publish the fork myself (of course while adhering to the license restrictions).


evrardjp:
This indeed has more value than dropping the molecule testing, which starts to annoy me in the long run. Would you be okay to help maintain the build from source ?  Else I am quite reluctant to add the build from source, as the PPA option (or other repos in other distros) is generally good enough.

I need to dig deeper into this. @zerwes  do you have an opinion on the build from source?

Salvoxia:
Thank your for responding!
I very much understand your hesitation.  
I have to admit I just very recently dipped my toes into backporting packages (truth be told, I only learned about that practice the other week.. yeah, I know..). While this has its own challenges, I think at the end of the day it does make more sense, especially if you want to stay compatible to future OS upgrades bringing new package versions.  
I have modified a couple of other Ansible roles in a similar way and it's working really well. I'll keep maintaining those, albeit that's just for my own amusement since nobody is using these anyway 😉 But I'll close this PR in favor of the solutions you proposed.

On a side note: The modifications I did for getting the Molecule tests up and running are mostly independent of the "compile from source" feature. Would you like/accept a PR for that only? I'm not sure what issues you've had with Molecule testing and systemd in the past. The roles I'm maintaining almost all rely on systemd services as well, and once the test is set up correctly it works reliably (or if it doesn't it's not systemd's fault).

